PR #19, opened 2018-05-24
added basis tranformation from ra/dec to za/az basis

zacharymartinot:
This adds the computation of the basis transformation to compute the coherency matrix in the local basis of the (Zenith-angle, Azimuth) spherical coordinates, assuming that the input coherency_radec matrix is defined in the RA/Dec basis according to the IAU convention: https://aas.aanda.org/articles/aas/pdf/1996/07/dst6557.pdf

### DIFF
--- a/pyuvsim/spherical_coordinates_basis_transformation.py
+++ b/pyuvsim/spherical_coordinates_basis_transformation.py
@@ -1,0 +1,70 @@
+import numpy as np
+
+def r_hat(theta,phi):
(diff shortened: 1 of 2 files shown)


bhazelton:
We need to figure out when this should be called (maybe not at zenith) and make sure it's being used in the appropriate tests.

bhazelton:
This branch also needs to be rebased on master.

jaguirre:
Closing because this was transferred to a new branch.

PR #30, opened 2018-05-29
Parameter reading/writing tools

### LINKED ISSUE #25: Enable configuration via yaml
Currently simulations are based on input files.  

The proposed yaml inputs:
simparams.yml : the main file input at the command line
telescopeparams.yml : things relevant to the observatory (including beams with IDs referenceable in antpos.csv)
antpos.csv : list of antenna positions (names, numbers, beam ids, and ENU positions)

### LINKED ISSUE #26: Import file to configuration YAML
Extract the relevant simulation parameters from a data file to a YAML file suitable for input to the simulation.

aelanman:
The param_reader branch has working functions for simulating from parameter files. It does this by generating a uvdata object based on parameters (and accepts several different parameter combinations for frequencies/time specification).

The data directory contains a beamfits file and several example configuration yamls. I wanted to make sure we were reading in a beamfits file with the parameter parsers (not just Nic's CST files), but doing so means we have to carry around a beamfits on the repo. Alternatively, we could write something that looks for the beamfits and, failing to find it, generates it in place. I'm open to suggestions if we think it'll be a problem to carry a beamfits around.

The unittests do two things. One unittest checks that reading in parameter files with different parameter configurations (ie., different ways of specifying the same thing) all return the same task list as an equivalent uvfits file read. The second unittest checks that the loop from uvfits -> parameter dictionary ->uvdata object -> uvfits file ->parameter dictionary works (comparing the first and last parameter dictionaries). I needed to run these tests off the 10time10channel test file because I had to test that the channel structure is generated correctly. There are some very small deviations in the results (of order milliseconds or millihertz), so I replaced the equalities with isclose() in the \_\_eq\_\_ functions. We should decide if these deviations are acceptable and if so what are the appropriate tolerances.

Fixes #25.  Fixes #26.


pyuvsim/data/test_config/28mEWbl_10time_10chan.yaml
```diff
@@ -0,0 +1,5 @@
+Nants: 2
```
bhazelton: Why do we want Nants here? I would think that should be dictated by the number of rows in the layout csv. That way you can use different layouts with the same telescope file.

pyuvsim/simsetup.py
```diff
@@ -0,0 +1,388 @@
...
+    uv_obj.Nbls = bls.size
+    uv_obj.Nblts = uv_obj.Nbls * uv_obj.Ntimes
+
+    uv_obj.ant_1_array, uv_obj.ant_2_array = \
+        uv_obj.baseline_to_antnums(uv_obj.baseline_array)
+
+    return uv_obj, beam_list, beam_ids
+
+
+def uvfits_to_telescope_config(file_in, beam_filepath, layout_csv_name=None,
```
bhazelton: Why is this specific to uvfits? Should we change it to take a UVData object so that it can be based on any file type?

pyuvsim/simsetup.py
```diff
@@ -0,0 +1,388 @@
...
+        yaml.dump(yaml_dict, yfile, default_flow_style=False)
+
+    print('Path: {}, telescope_config: {}, layout: {}'.format(path, telescope_config_name,
+                                                      layout_csv_name))
+
+    if return_names:
+        return path, telescope_config_name, layout_csv_name
+
+
+def uvfits_to_config_file(file_in, config_filename=None, keep_path=False, telescope_config_path='',
```
bhazelton: same question about uvfits vs UVData here.

pyuvsim/simsetup.py
```diff
@@ -0,0 +1,388 @@
...
+    dt = np.format_parser(['a10', 'i4', 'i4', 'f8', 'f8', 'f8'],
+                          ['name', 'number', 'beamid', 'e', 'n', 'u'], header)
+
+    return np.genfromtxt(layout_csv, autostrip=True, skip_header=1,
+                         dtype=dt.dtype)
+
+
+def initialize_uvdata_from_params(param_dict):
+    """
+        Construct a uvdata object from parameters in a valid yaml file.
```
bhazelton: this doc string should indicate that you're also constructing the UVBeam objects.

pyuvsim/simsetup.py
```diff
@@ -0,0 +1,388 @@
...
+
+    param_dict.update(telparam)
+    E, N, U = ant_layout['e'], ant_layout['n'], ant_layout['u']
+    antnames = ant_layout['name']
+    beam_ids = ant_layout['beamid']
+    beam_list = []
+    for beamID in np.unique(beam_ids):
+        path = telparam['beam_paths'][beamID]
+        uvb = UVBeam()
+        uvb.read_beamfits(path)
```
bhazelton: I think this means that you cannot pass Nic's CST beam files to this code. Is that what we want?

pyuvsim/simsetup.py
aelanman: As it's currently written, it won't generate the UVBeam object from a CST file. It's designed to read in a beamfits file. If you prefer, I can make it check whether the file path is to a txt or .beamfits file and then either generate the UVBeam or read it in.

pyuvsim/data/test_config/28mEWbl_10time_10chan.yaml
aelanman: I wrote it originally thinking it would use Nants as an extra constraint. I'll switch to using the number of rows instead.

I just noticed though, while getting my nosetests to work on Oscar, that the config files require the full paths to the telescope parameters and layout files. At the moment, they're set to the paths on my laptop. Perhaps I can have it default to checking the path relative to the location of the config file?

pyuvsim/simsetup.py
aelanman: Good idea. I'll switch that out.

jpober:
Fixes #25 and fixes #26.

PR #33, opened 2018-06-19
Progress steps

aelanman:
Supplements the progbar functionality with a "steps" option when pyuvsim is run as a batch job. The progress bar doesn't look nice when written to log files instead of the stdout, so this lets you write progress to files. It checks for an environmental variable "PYUVSIM_BATCH_JOB" to decide which to use.

### DIFF
--- a/pyuvsim/utils.py
+++ b/pyuvsim/utils.py
@@ -1,6 +1,7 @@
 import time as pytime
 import sys
 
+
 class progsteps:
     """
         Similar to progress bar. Prints a percentage of completion.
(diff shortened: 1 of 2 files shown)


aelanman:
Is it not passing coveralls because I didn't write a unit test for the progsteps class?

bhazelton:
I'm not sure, I'm looking at the coveralls settings and there's not a set COVERAGE THRESHOLD FOR FAILURE or COVERAGE DECREASE THRESHOLD FOR FAILURE, so I'm not sure why it's failing. We probably do want to set those eventually...

bhazelton:
I'm inclined to merge this and make an issue about increasing test coverage. What do you think @jpober?

jpober:
Sounds good to me @bhazelton

PR #40, opened 2018-07-12
remove epoch, clarify that source ra/dec are in J2000

### LINKED ISSUE #31: Only use astropy's 'icrs' frame if the epoch is J2000. Otherwise use FK5
Right now we are using 'icrs' with an epoch which doesn't do anything (an icrs frame doesn't take an epoch).

bhazelton:
Also adds pep8 checking to Travis and fixes a bunch of pep8 problems.

fixes #31

### DIFF
--- a/.coveragerc
+++ b/.coveragerc
@@ -1,0 +1,5 @@
+[run]
+omit = */tests/*
+
+[report]
+omit = */tests/*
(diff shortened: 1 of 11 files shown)


bhazelton:
Ok, I think this is ready to go.

bhazelton:
The coverage decrease is because it's no longer counting the tests in the coverage (which it shouldn't have been counting in the first place)

PR #53, opened 2018-07-19
Unittest multiple beams

aelanman:
Makes the following changes:

- Redefines the test config files to use a triangle antenna layout with a different beam model for each antenna (HERA, tophat, and gaussian).
- Updates the 28m triangle test file with the new phasing.
- Replaces the \_\_cmp\_\_ functions with the python3 compatible comparison operators, for sorting tasks.
- Updates simsetup unit tests to use multiple beam models.


pyuvsim/tests/test_simsetup.py
```diff
@@ -149,3 +160,7 @@ def test_point_catalog_reader():
         nt.assert_true(src.dec.deg in catalog_table['dec_j2000'])
         nt.assert_true(src.stokes[0] in catalog_table['flux_density_I'])
         nt.assert_true(src.freq.to("Hz").value in catalog_table['frequency'])
+
+
+if __name__ == '__main__':
```
bhazelton: did you mean to leave this in?

bhazelton:
I don't see a `28m_triangle_10time_10chan.yaml` file in this PR or in the existing data files.

pyuvsim/tests/test_simsetup.py
aelanman: Ah, no ... sorry.

aelanman:
I think I left that out by mistake when adding to the commits.

There are other significant problems at the moment. I did all this without the latest pyuvdata which is why it's not building with Travis. I'm working to fix that now.

aelanman:
@bhazelton This should be ready for review.

PR #74, opened 2018-07-23
Redo mock catalog and wrappers

aelanman:
I didn't like that all of the keywords for "create_mock_catalog" had to be keywords for "run_pyuvsim" as well, so I've wrapped them up in a dictionary. The accepted keywords are still properly documented in the create_mock_catalog docstring (ie., no "kwargs"), but run_pyuvsim has keywords that are relevant to it now.

I've also deleted the "run_serial_uvsim" function. It's no longer used by anything except a single unit test (also deleted), which just checks that it runs. We have a unit test for "run_uvsim" already.

Lastly, the wrappers run_pyuvsim and run_param_pyuvsim have been updated to work with the new setup.

### DIFF
--- a/pyuvsim/simsetup.py
+++ b/pyuvsim/simsetup.py
@@ -15,6 +15,36 @@ from astropy.coordinates import Angle
 # and for generating parameter files from uvfits files
 
 
+def write_uvfits(uv_obj, param_dict):
+    """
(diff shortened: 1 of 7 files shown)


aelanman:
I will need the functionality included here to run the reference sims. In particular, this includes the means to run GLEAM simulations with our wrappers.

PR #76, opened 2018-07-24
Version fix

### LINKED ISSUE #73: Unstable behavior of subprocess in MPI
https://github.com/RadioAstronomySoftwareGroup/pyuvsim/blob/9b28a3da13bd825ed60fff3dca3f795256c6cdb2/pyuvsim/version.py#L22-L25

These lines occasionally segfault, but each time they're executed I get the following warning from OpenMPI:

	--------------------------------------------------------------------------
	A process has executed an operation involving a call to the
	"fork()" system call to create a child process.  Open MPI is currently
	operating in a condition that could result in memory corruption or
	other system errors; your job may hang, crash, or produce silent
	data corruption.  The use of fork() (or system() or other calls that
	create child processes) is strongly discouraged.
	
	The process that invoked fork was:
	
	  Local host:          [[13586,1],0] (PID 29335)
	
	If you are *absolutely sure* that your application will successfully
	and correctly survive a call to fork(), you may disable this warning
	by setting the mpi_warn_on_fork MCA parameter to 0.
	--------------------------------------------------------------------------
	

Some googling suggests that we should replace the subprocess calls with a call to `MPI.COMM_SELF.Spawn`, but I'm not exactly sure how to do that.

This seems to be an Oscar-specific issue. I don't see this when running on my laptop.

aelanman:
Uses os.popen instead of the subprocess module in version.py to get the git information. This *should* be safe, since we're not passing user data to the system call, and it does seem to work fine under MPI on Oscar.

Fixes #73

### DIFF
--- a/pyuvsim/version.py
+++ b/pyuvsim/version.py
@@ -24,10 +24,6 @@ def construct_version_info():
             data = os.popen(" ".join(argv)).read()
         else:
             data = ""
-#        if capture_stderr:
-#            data = subprocess.check_output(argv, stderr=subprocess.STDOUT)
-#        else:
-#            data = subprocess.check_output(argv)
 
         data = data.strip()
 
--- a/pyuvsim/version.py
+++ b/pyuvsim/version.py
@@ -8,6 +8,9 @@ import os
 import six
 import subprocess
 import json
+from mpi4py import MPI
+
+rank = MPI.COMM_WORLD.Get_rank()
 
 
 def construct_version_info():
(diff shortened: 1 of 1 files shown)


jpober:
@bhazelton Do you have to review/think about this change?  Right now, Oscar is the only machine we have running simulations, and we need this to reliably run on Oscar.  But it's not clear it's the "right" thing to do generally.  Perhaps we should merge it and revisit when we expand to other clusters?

bhazelton:
I have no insight here, I think we can do as you suggest @jpober

jpober:
Sounds like a plan for now

PR #79, opened 2018-07-26
Make baseline-lite array and two-source catalog, organize ref_sim files

### LINKED ISSUE #63: Create 10 source test pattern
Make an "R" (tab-delimited catalog).

### LINKED ISSUE #64: Create 2 source test pattern
2 sources 160 degrees apart in RA at same declination.

### LINKED ISSUE #65: Define baseline-lite pattern
4 antennas (10 baselines w/ autocorrelations) for reference simulations.

aelanman:
Adds the required catalogs and array layouts. I've also copied the lighter catalogs to the reference_simulations directory to better preserve a record of what was used and make it easier to share later.

I assume issue 67 is the "letters on the horizon" catalog. That is not included here.

Fixes #65, Fixes #64, Fixes #63


aelanman:
The baseline-lite layout is a right triangle of antennas (50 meters EW and NS) plus an antenna that is sqrt(2) meters off from the midpoint of the hypotenuse.

PR #84, opened 2018-08-02
Fix version

bhazelton:
make versions work properly, add tests.

### DIFF
--- a/pyuvsim/version.py
+++ b/pyuvsim/version.py
@@ -76,7 +76,7 @@ git_branch = version_info['git_branch']
 
 
 def main():
-    print("Rank = {0}".format(rank))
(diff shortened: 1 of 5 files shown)


pyuvsim/version.py
```diff
@@ -76,7 +76,6 @@ def unicode_to_str(u):
 
 
 def main():
-    print("Rank = {0}".format(rank))
```
bhazelton: @aelanman I took this line out to make the tests pass, is it needed for some reason?

aelanman: It should be fine. I think I just put that there to track what each rank was printing.

bhazelton:
I think this is ready to go if there's no objections.

PR #86, opened 2018-08-03
Reference sim changes, round 2

aelanman:
Changes:

1. Fix location of the MWA88_nocore to the HERA site.
2. Convert catalog RAs to decimal degrees, and include units in brackets in the header line.
3. Rename tophat beam to 'uniform' for clarity.
4. Use a 11.4 degree FHWM gaussian beam when a gaussian is called for.
5. Fix bug related to scattering UVBeam objects.

I'm rerunning the reference simulations now. The HERA beam ones are still up against the int overflow bug, but I may be able to temporarily circumvent that by just passing along the beam filename and having each rank load it separately.


pyuvsim/uvsim.py
```diff
@@ -1010,6 +1010,13 @@ def run_uvsim(input_uv, beam_list, beam_dict=None, catalog_file=None,
     if rank == 0:
         print("Tasks Received. Begin Calculations.")
         stdout.flush()
+
+    for i, bm in enumerate(local_task_list[0].telescope.beam_list):
+        if isinstance(bm, UVBeam):
+            uvb = UVBeam()
+            uvb = bm
+            local_task_list[0].telescope.beam_list[i] = bm
```
bhazelton: Maybe add a comment explaining what's happening here and why.

bhazelton:
Looks good to me. Should we wait until the sims finish to make sure everything worked as intended?

pyuvsim/uvsim.py
aelanman: Good idea. The cluster seems busy at the moment... the jobs are still pending and the estimated start time isn't for a couple days. In my experience they'll run much sooner than that.

aelanman:
Good idea. The cluster seems busy at the moment... the jobs are still pending and the estimated start time isn't for a couple days. In my experience they'll run much sooner than that.

PR #89, opened 2018-08-09
Herabeam workaround

aelanman:
This changes things up so that the beam objects are not instantiated until after the scatter, avoiding the MPI int overflow bug. It was branched off of reference_sim_changes.

### DIFF
--- a/pyuvsim/simsetup.py
+++ b/pyuvsim/simsetup.py
@@ -20,6 +20,33 @@ from .analyticbeam import AnalyticBeam
 from .mpi import rank
 from pyuvsim.data import DATA_PATH as SIM_DATA_PATH
 
+# Utilities for setting up simulations for parameter files,
+# and for generating parameter files from uvfits files
+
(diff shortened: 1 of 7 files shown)


pyuvsim/simsetup.py
```diff
@@ -207,28 +229,14 @@ def initialize_uvdata_from_params(obs_params):
...
-            path = beam_model   # beam_model = path to beamfits
-        uvb.read_beamfits(path)
-        beam_list.append(uvb)
+        if beam_model == 'gaussian':
+            try:
+                sigma = telparam['sigma']
+                beam_model = beam_model + '_' + str(sigma)
+            except KeyError:
+                print("Missing sigma for gaussian beam.")
+                raise err
```
bhazelton: why not just raise a value error here? In general, I think we should avoid print statements for warnings and errors -- instead use the warnings package and raise the appropriate error types.

aelanman:
Out of date. Issues will be solved by local_task_generation_v2

PR #90, opened 2018-08-11
Fix airy simsetup

anguta:
Airy beams in setup.py where broken since they were not listed as an analytic option in `initialize_uvdata_from_params`


pyuvsim/GIT_INFO
```diff
@@ -0,0 +1 @@
+["", "", "", ""]
```
bhazelton: This should not be added to the repo! If it's showing up it should be added to the .gitignore.

anguta:
One question -- are the analytic beam functions referring to power or e-field responses? As written, the airy beam is the power response but my interpretation of the code indicates that I should have written the voltage response.

bhazelton:
Good question. The analytic beam functions refer to Efield beams not power beams. More documentation on the sigma would be good!

bhazelton:
@anguta Sorry, but we needed to merge the cleanup branch and the airy beam tests weren't passing, even when we tried to pull in these changes to the cleanup branch. So can you either rebase this or just make a new branch with these fixes (which might be easier).

bhazelton:
Looks good to me! @aelanman are you happy?

PR #102, opened 2018-08-31
fix for tests failing on master

bhazelton:


bhazelton:
@aelanman the tests are passing on Travis now, but not on my local machine. Can you check that this branch works for you on Oscar?

The problem when I run locally is in  test_mpi_uvsim.py:test_run_param_uvsim the data_arrays don't match -- `nt.assert_equal(uv_new, uv_ref)` fails. It looks like that test is comparing against a saved file. Is that right? I don't understand why it's working on Travis but not locally...

bhazelton:
Note that I increased the checking on the `config_path` variable. But from the code, I really don't understand why that variable is present in _any_ of the files -- it looks like it should be set by the location of the file, not by something saved in the file (which can of course get out of date if the file is moved). I also don't see that field documented anywhere in the docs, which makes me think it's a mistake. Should we remove that line from all the test files that have it (some don't)?

aelanman:
That is worrying... if the data arrays aren't matching on one machine but are on another, that sounds like there's a difference in the quantities being computed (since it's the same saved file). I compare against a saved file

I included the `config_path` keyword because I originally wrote the `initialize_uvdata_from_params` function to accept a params dictionary as its input (not the filename), so it would have no way of knowing where to find the telescope config and layout files unless the dictionary carries in the file path. I'm not sure what the best way forward is. Should we switch to having `initialize_uvdata_from_params`  run solely from a filename?

Part of the problem with including MPI in the unit tests is that you can call MPI.Init() and MPI.Finalize() exactly once per python session. I managed to avoid calling MPI.Init() multiple times, but there's still the problem that if a test using `subprocess` is called after a test using MPI, then the MPI environment is still initialized and I'll get a fork warning and a segmentation fault. If I run each test file individually, they pass just fine, but if I call them all at once (ie., just called `nosetests`) then it gives me the fork warning and occasionally fails.

All that being said, these changes look fine and the tests do pass on Oscar, if inconsistently.

bhazelton:
@aelanman I think the inclusion of the `config_path` in the dict is appropriate. If the dict is read from a file then `initialize_uvdata_from_params` fills in that keyword using the file location. If the dict is passed to that function, then the `config_path` keyword needs to be added before passing the dict. Note that you were effectively doing in the test -- overwriting the incorrect value from the file -- but it was a bit non-intuitive, so I cleaned it up a little by setting that path explicitly to the location of the other config files.

That being said, I'm comfortable with changing the `initialize_uvdata_from_params` function to just take a file if that is the pattern we're essentially always going to use.

bhazelton:
To clarify my last comment, I think that the `config_path` should be in the dict but not in the file.

bhazelton:
@aelanman I think this is ready for your review.

bhazelton:
I haven't had time to investigate further. But it's working for you and on Travis and these changes fix some of my problems, so I'd like to merge.

PR #103, opened 2018-09-04
Adding compatibility with python 3

### LINKED ISSUE #39: Make python 3 compatible
Should convert code to be python 3 compatible.

bhazelton:
closes #39

### DIFF
--- a/pyuvsim/tests/test_mpi_uvsim.py
+++ b/pyuvsim/tests/test_mpi_uvsim.py
@@ -10,8 +10,6 @@ import yaml
 import nose.tools as nt
 from mpi4py import MPI
 
-from astropy import time
-
 from pyuvdata import UVBeam, UVData
 from pyuvdata.data import DATA_PATH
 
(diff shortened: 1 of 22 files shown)


mkolopanis:
This closes #39 but does that need to be in the top level comment?

bhazelton:
yes, the closing text needs to be in the description, I'll add it.

pyuvsim/tests/test_mpi_uvsim.py
```diff
@@ -62,8 +64,16 @@ def test_run_param_uvsim():
     uv_new = UVData()
     uv_new.read_uvfits(tempfilename)
     os.remove(tempfilename)
+
+    time_obj = time.Time(uv_new.time_array[0], format='jd')
+    uv_new.unphase_to_drift(use_ant_pos=True)
+    uv_new.phase_to_time(time_obj)
```
bhazelton: On my machine, I don't need this `phase_to_time(time_obj)` line, the objects match with just the `unphase_to_drift(use_ant_pos=True)` added. What was the problem you were seeing without the `phase_to_time`?

bhazelton: I see it. It's actually that the uv_new object in the lower block of code (after the votable stuff)  is entirely new -- read from a uvfits file (so it's a phased object), while it's being compared to the uv_ref that was already unphased. Adding `uv_new.unphase_to_drift(use_ant_pos=True)` before the final assert fixes it.

mkolopanis: ah I missed that second block. I will add that change.

dannyjacobs:
Adds python3 compatibility
Fixes coverall incorrectly reports coverage decreasing

PR #110, opened 2018-09-14
Redesign mpi imports and ensure MPI.Init is only run once

aelanman:
As it turns out, running MPI.Init() twice doesn't throw an exception but instead just causes an MPI error. I'm now aware of the MPI.Is_initialized() function, which makes this a lot simpler.

Unfortunately, I had to put a module level import _after_ setting mpi4py.rc.initialize=False. This breaks PEP8 rules, so I've changed the travis pycodestyle call to ignore E402 when checking `mpi.py`. This is really a flaw in mpi4py's design, which by default initializes MPI when imported...

### DIFF
--- a/pyuvsim/tests/test_mpi_uvsim.py
+++ b/pyuvsim/tests/test_mpi_uvsim.py
@@ -28,6 +28,7 @@ singlesource_txt = os.path.join(SIM_DATA_PATH, 'single_source.txt')
 
 
 def test_run_uvsim():
+    print('test_run_uvsim')
     hera_uv = UVData()
     hera_uv.read_uvfits(EW_uvfits_file)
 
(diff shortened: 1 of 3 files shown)


bhazelton:
You can tell pycodestyle to ignore specific lines by adding this comment at the end of the line: #noqa

pyuvsim/tests/test_mpi_uvsim.py
```diff
@@ -28,6 +28,7 @@
 
 
 def test_run_uvsim():
+    print('test_run_uvsim')
```
bhazelton: Did you want these print statements to stay in?

pyuvsim/tests/test_mpi_uvsim.py
```diff
@@ -48,6 +49,7 @@ def test_run_uvsim():
 
 
 def test_run_param_uvsim():
+    print('test_run_param_uvsim')
```
bhazelton: ditto

pyuvsim/tests/test_mpi_uvsim.py
```diff
@@ -86,6 +88,7 @@ def test_run_param_uvsim():
 
 
 def test_mpi_funcs():
+    print('test_mpi_funcs')
```
bhazelton: ditto

aelanman:
Argh... sorry. I put those in to look for where the mpi error arose.

PR #112, opened 2018-09-21
Start cleaning up test code

### LINKED ISSUE #51: Clean up
Code needs a thorough read through and reorganization before things get further out of hand.

### LINKED ISSUE #108: Handle warnings in tests
Something like what's in pyuvdata that catches the warnings and checks that they are as expected would be ideal. It's hard to read the output of nosetests right now.

bhazelton:
Still to do:
 - ~figure out what's going on with `test_simsetup.py: test_setup_airy`~
 - ~figure out whether to keep `test_uvsim.py:test_single_offzenith_source_miriad`~
 - ~catch warnings~
 - ~improve the quality of the testing of the mock catalog creation (to include locations of sources, not just numbers of them)~
- ~move toward one assert per test~ (I think this is good enough...)

fixes #51 
fixes #108


bhazelton:
@anguta can you look at the `test_simsetup.py: test_setup_airy` function? The first line of that function seems not to be actually testing anything -- there should be some sort of assert associated with it.

aelanman:
The problem with `test_single_offzenith_source_miriad` is that the antenna positions table in the testfile (in pyuvdata's data directory) doesn't match the actual antenna layout, so the ENU antpos array is nonsensical and the baselines it generates don't match what's expected from the uvw array.

Since this test essentially repeats `test_visibility_single_zenith_source_uvdata` except with a different file, I propose removing it.

bhazelton:
Ok, thanks for figuring that out @aelanman, I took it out.

aelanman:
No problem. I guess the last thing is to understand the test_setup_airy function?

I've looked at it more closely. I'm pretty sure it's just testing that trying to load an airy beam from a parameter file without a diameter specified will raise an error. This is already accomplished in `test_param_reader`, so I'm inclined to remove @anguta 's test and its associated config files.

bhazelton:
ok @jpober or @dannyjacobs I think this PR is ready for review.

PR #113, opened 2018-09-25
Use alt az

### LINKED ISSUE #109: fix azimuth interface with UVBeam
Currently when UVBeam is called, azimuth angle is not being converted from astropy frame to the UVBeam/CST/Radio Astronomy convention.

bhazelton:
fixes #109

### DIFF
--- a/pyuvsim/analyticbeam.py
+++ b/pyuvsim/analyticbeam.py
@@ -28,7 +28,7 @@ class AnalyticBeam(object):
 
     @profile
     def interp(self, az_array, za_array, freq_array):
-        # (Naxes_vec, Nspws, Nfeeds or Npols, freq_array.size, az_array.size)
+        # The azimuth here has the UVBeam convention: North of East(East=0, North=90)
 
         if self.type == 'uniform':
             interp_data = np.zeros((2, 1, 2, freq_array.size, az_array.size), dtype=np.float)
--- a/pyuvsim/analyticbeam.py
+++ b/pyuvsim/analyticbeam.py
@@ -28,8 +28,24 @@ class AnalyticBeam(object):
 
     @profile
     def interp(self, az_array, za_array, freq_array):
-        # The azimuth here has the UVBeam convention: North of East(East=0, North=90)
+        """
+        Interpolate beam to given az, za locations (in radians).
 
+        Mirrors the interp function of UVBeam objects.
+        Args:
+            az_array: az values to interpolate to in radians (same length as za_array)
+                The azimuth here has the UVBeam convention: North of East(East=0, North=pi/2)
+            za_array: za values to interpolate to in radians (same length as az_array)
+            freq_array: frequency values to interpolate to
+
+        Returns:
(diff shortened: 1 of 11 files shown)


pyuvsim/analyticbeam.py
```diff
@@ -28,7 +28,7 @@ def peak_normalize(self):
 
     @profile
     def interp(self, az_array, za_array, freq_array):
-        # (Naxes_vec, Nspws, Nfeeds or Npols, freq_array.size, az_array.size)
+        # The azimuth here has the UVBeam convention: North of East(East=0, North=90)
```
dannyjacobs: az_array = azimuth in degrees from East rotating towards the North. East=0, North = 90
za_array = zenith angle in degrees away from zenith. straight up=0, horizon=90
freq_array = frequencies in Hz

Three arrays may be in arbitrary shape, but shapes of all three must be the same.

pyuvsim/source.py
```diff
@@ -116,16 +119,16 @@ def coherency_calc(self, time, telescope_location):
...
-        calculate the azimuth & zenith angle for this source at a time & location
+        calculate the altitude & azimuth for this source at a time & location
 
         Args:
             time: astropy Time object
             telescope_location: astropy EarthLocation object
 
         Returns:
-            (azimuth, zenith_angle) in radians
+            (altitude, azimuth) in radians
```
dannyjacobs: Looks like we switch between degrees and radians.  Should we try to be consistent?

pyuvsim/utils.py
```diff
@@ -75,3 +76,65 @@ def cirs_to_tee_ra(cirs_ra, time):
...
+    """
+    Convert from astropy altaz convention to UVBeam az/za convention.
+
+    Args:
+        zenith_angle: in radians
+        azimuth: in radians in uvbeam convention: North of East(East=0, North=90)
+
+    Returns:
+        altitude in radians
+        azimuth in radians in astropy convention: East of North (N=0, E=90 degrees)
```
dannyjacobs: Notes in degrees, units in radians.

Returns:
        altitude in radians
        azimuth in radians in astropy convention: East of North (N=0, E=pi/2 degrees)

pyuvsim/analyticbeam.py
bhazelton: I added a full docstring.

pyuvsim/source.py
bhazelton: I think we've consistently used radians, although some comments reference degrees which I will change.

bhazelton: we do use degrees in some testing, but I think that's ok as they're always converted to radians before calling the functions.

pyuvsim/utils.py
bhazelton: fixed.

bhazelton:
Ok @dannyjacobs I think this is ready for review.

pyuvsim/source.py
dannyjacobs: yeah, thats what it looks like to me too.  Seems ok.

PR #114, opened 2018-09-27
Docstring cleanup

aelanman:
Changes to the docstrings on classes and functions, and an update to the sphinx documentation.

### DIFF
--- a/pyuvsim/analyticbeam.py
+++ b/pyuvsim/analyticbeam.py
@@ -53,23 +53,7 @@ class AnalyticBeam(object):
                 if az/za_arrays are not passed), shape: (Naxes_vec, Ncomponents_vec,
                 Npixels/(Naxis1, Naxis2) or az_array.size if az/za_arrays are passed)
         """
-        # (Naxes_vec, Nspws, Nfeeds or Npols, freq_array.size, az_array.size)
 
-        Mirrors the interp function of UVBeam objects.
-        Args:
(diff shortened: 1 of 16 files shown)


aelanman:
This should be ready now.

pyuvsim/analyticbeam.py
```diff
@@ -29,23 +37,23 @@ def peak_normalize(self):
...
 
-        Mirrors the interp function of UVBeam objects.
         Args:
-            az_array: az values to interpolate to in radians (same length as za_array)
-                The azimuth here has the UVBeam convention: North of East(East=0, North=pi/2)
-            za_array: za values to interpolate to in radians (same length as az_array)
-            freq_array: frequency values to interpolate to
+            az_array: az values to evaluate (same length as za_array)
+            za_array: za values to evaluate (same length as az_array)
+            freq_array: frequency values
```
bhazelton: I'd like to keep at least the args entries the way they were before.

aelanman: I restored the args text as before, except I replaced "interpolate" with "evaluate." A minor nitpick, but since this is an analytic beam we're not interpolating but actually returning an exact function value. Is that okay?

bhazelton: Yes, thanks! I just didn’t want to lose the units and convention notes

bhazelton:
I’m comfortable with you merging this now @aelanman

PR #115, opened 2018-10-03
Local task generation v2

### LINKED ISSUE #54: Delete task list after scatter
Potential way to make more RAM available.  Need to explore.

### LINKED ISSUE #56: Make task calculation local to nodes
Proposed solution for reducing memory overhead in rank 0 process during tasklist creation and scatter. Instead broadcast the simulation parameters and make the nodes calculate their tasks on the fly using their rank and the number of tasks.

This will merge nicely with the recalculation minimization optimizations Adam L. was playing with on the sorttasks branch.

### LINKED ISSUE #93: re-merge herabeam workaround with master post cleanup
there will be conflicts with master after we merge in the cleanup branch

aelanman:
I've remade the `local_task_generation` branch, since it was easier than rebasing. This implements a few tricks that speed things up a little:

1. Reuse alt/az position calculations for all baselines and frequencies per given source and time.
2. Generate tasks on each processor, instead of making a big task list and scattering it. It does this using the rank and Npus to figure out which tasks to complete, and 
3. Since we're no longer scattering them, we can start with times and frequencies as astropy Time and Quantity objects. It's generally faster to do that conversion once on an array than to call it multiple times on individual floats (microseconds vs. milliseconds), so it does that.
4. ~~(To be written)~~ Use the `reuse_spline` feature of UVBeam to speed up beam interpolation steps. This should make HERA-beam simulations feasible.
5. (10/18/18) Broadcast strings specifying beams to make/load, and then make beam objects on rank. This is the 'herabeam_workaround'.

Some trade-offs:
1. The initial UVData object has to be broadcast to each processor. Each needs to know all the times, frequencies, and baselines, and this is the simplest way to do that. (10/24/18) The catalog and beam lists also need to be copied onto each rank, which quickly increases memory usage with Npus. Shared memory and/or a cleverer scattering system will be needed to resolve this.
2. Repetition is (theoretically) reduced on each processor by reusing previously-calculated quantities, but globally there will still be some repetition. There may, therefore, be a more complicated relationship between Npus and runtime.

Fixes #56 , Fixes #93 , Fixes #54


pyuvsim/source.py
```diff
@@ -157,8 +166,8 @@ def pos_lmn(self, time, telescope_location):
             (l, m, n) direction cosine values
         """
         # calculate direction cosines of source at current time and array location
-        if self.alt_az is None:
-            self.alt_az_calc(time, telescope_location)
+        # Will only do the calculation if time has changed
+        self.alt_az_calc(time, telescope_location)
```
bhazelton: I don't think this comment is true currently. I think for it to be true you need to call your new function `get_alt_az`.

pyuvsim/uvsim.py
```diff
@@ -276,101 +272,39 @@ def initialize_uvdata(uvtask_list, source_list_name, uvdata_file=None,
...
-    uv_obj.set_lsts_from_time_array()
-    uv_obj.zenith_ra = uv_obj.lst_array
-    uv_obj.zenith_dec = np.repeat(uv_obj.telescope_location_lat_lon_alt[0], uv_obj.Nblts)  # Latitude
-    uv_obj.object_name = 'zenith'
-    uv_obj.set_drift()
-    uv_obj.vis_units = 'Jy'
-    uv_obj.polarization_array = np.array([-5, -6, -7, -8])
-    uv_obj.spw_array = np.array([0])
-    uv_obj.freq_array = np.array([freqs])
+        uv_obj.integration_time = (np.ones_like(uv_obj.time_array, dtype=np.float64))
```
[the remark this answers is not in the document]
bhazelton: Note also that this line is currently not covered by tests.

pyuvsim/uvsim.py
```diff
@@ -136,101 +140,94 @@ def make_visibility(self):
...
-def uvdata_to_task_list(input_uv, sources, beam_list, beam_dict=None):
+def uvdata_to_task_iter(task_ids, input_uv, catalog, beam_list, beam_dict):
     """
-    Create task list from pyuvdata compatible input file.
+    Generate local tasks, reusing quantities where possible.
 
     Args:
+        task_ids (numpy.ndarray of ints): Task index in the full flattened meshgrid of parameters.
         input_uv (UVData): UVData object to use
         sources: array of Source objects
```
bhazelton: this parameter is now called `catalog` not sources.

pyuvsim/uvsim.py
```diff
@@ -276,101 +272,39 @@ def initialize_uvdata(uvtask_list, source_list_name, uvdata_file=None,
...
-    antnums = task_antnums[ant_indices]
-    antpos = task_antpos[ant_indices]
-
-    freqs = np.unique(task_freqs)
-
-    uv_obj = UVData()
+    history += ' Npus = ' + str(mpi.Npus) + '.'
 
-    # add pyuvdata version info
-    history += uv_obj.pyuvdata_version_str
```
bhazelton: Is the input uv guaranteed to already have both the pyuvdata and pyuvsim version strings?

pyuvsim/uvsim.py
```diff
@@ -276,101 +272,39 @@ def initialize_uvdata(uvtask_list, source_list_name, uvdata_file=None,
...
-    uv_obj.antenna_numbers = antnums
-    antpos_ecef = uvutils.ECEF_from_ENU(antpos, *uv_obj.telescope_location_lat_lon_alt) - uv_obj.telescope_location
-    uv_obj.antenna_positions = antpos_ecef
-    uv_obj.ant_1_array = np.array(ant_1_array)
-    uv_obj.ant_2_array = np.array(ant_2_array)
-    uv_obj.time_array = np.array(task_times)
-    uv_obj.uvw_array = np.array(task_uvw)
-    uv_obj.baseline_array = uv_obj.antnums_to_baseline(ant_1_array, ant_2_array)
-    uv_obj.Nbls = np.unique(uv_obj.baseline_array).size
+    uv_obj.object_name = 'zenith'
```
bhazelton: hard-coding this to zenith strikes me as a little odd. At a minimum I think we should allow it to be set in some way to something meaningful.

pyuvsim/uvsim.py
```diff
@@ -276,101 +272,39 @@ def initialize_uvdata(uvtask_list, source_list_name, uvdata_file=None,
...
-    uv_obj.Npols = uv_obj.polarization_array.size
+    # add pyuvdata version info
+    history += uv_obj.pyuvdata_version_str
 
+    # Clear existing data, if any
     uv_obj.data_array = np.zeros((uv_obj.Nblts, uv_obj.Nspws, uv_obj.Nfreqs, uv_obj.Npols), dtype=np.complex)
     uv_obj.flag_array = np.zeros((uv_obj.Nblts, uv_obj.Nspws, uv_obj.Nfreqs, uv_obj.Npols), dtype=bool)
     uv_obj.nsample_array = np.ones_like(uv_obj.data_array, dtype=float)
     uv_obj.history = history
+    uv_obj.set_drift()
```
bhazelton: I think this is already done on line 280

pyuvsim/uvsim.py
aelanman: This is currently how the object name is set on master, but I agree it should have a more meaningful name. Maybe the catalog name?

bhazelton: Maybe catalog + something about the RA/Dec of zenith during the sim -- This is usually used in radio astronomy to indicate what object(s) or parts of the sky the observation was looking at. We could also allow people to set it in some way that was meaningful to them...

pyuvsim/uvsim.py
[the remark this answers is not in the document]
aelanman: Looks like I forgot to copy over the line immediately after, which multiplies by the np.diff of unique(time_array).

pyuvsim/source.py
aelanman: You're right... assuming this doesn't break things completely, it might mean that we'd actually get a further speedup, since it wasn't actually reusing previously-calculated alt/az values.

aelanman:
My last commit should take care of most of these things @bhazelton. I have it set the `object_name` to the <catalog name>_ra<ra.deg>_dec<dec.deg>, where the ra/dec are rounded to four decimal places. It currently includes the full file path for the catalog, which may not be desired...

If a user wants to specify a particular `object_name`, it can be passed through the parameter yaml along with any other valid UVData parameter. That might not be very clear from the documentation...

jpober:
Before you merge, can you push an update to the version number (0.2.0) and then tag a new release after the merge?

bhazelton:
@aelanman I think we should also maybe add a changelog. Our issue on that says it should start immediately after the first release

aelanman:
@bhazelton @jpober I've added a basic changelog and changed the version number, and I've raised an issue for the memory limitation.

PR #136, opened 2018-11-05
 Add support for selecting baselines to simulate

### LINKED ISSUE #59: Make it possible to turn off simulating autocorrelations

### LINKED ISSUE #62: Make it possible to set which baselines to simulate

bhazelton:
currently requires the `select_metadata` branch of pyuvdata. That Travis line should be reverted when that branch is merged into pyuvdata's master

fixes #59 
fixes #62

### DIFF
--- a/pyuvsim/simsetup.py
+++ b/pyuvsim/simsetup.py
@@ -542,11 +542,12 @@ def initialize_uvdata_from_params(obs_params):
     uv_obj.ant_1_array, uv_obj.ant_2_array = \
         uv_obj.baseline_to_antnums(uv_obj.baseline_array)
 
-    # add other required metadata to allow select on metadata to work
+    # add other required metadata to allow select to work without errors
+    # these will all be overwritten in uvsim.init_uvdata_out, so it's ok to hardcode them here
     uv_obj.polarization_array = np.array([-5, -6, -7, -8])
     uv_obj.set_lsts_from_time_array()
     uv_obj.set_uvws_from_antenna_positions()
-    uv_obj.history = ''  # this will get overwritten in uvsim.init_uvdata_out
+    uv_obj.history = ''
 
     # down select baselines (or anything that can be passed to pyuvdata's select method)
     # Note: cannot down select polarizations (including via ant_str or bls keywords)
--- a/pyuvsim/simsetup.py
+++ b/pyuvsim/simsetup.py
@@ -526,7 +526,9 @@ def initialize_uvdata_from_params(obs_params):
         param_dict['object_name'] = '{}_ra{:.4f}_dec{:.4f}'.format(param_dict['sources']['catalog'], src.ra.deg, src.dec.deg)
     uv_obj = UVData()
     for k in param_dict:
-        if hasattr(uv_obj, k):
+        # use the __iter__ function on UVData to get list of UVParameters on UVData
+        param_name = [getattr(uv_obj, param).name for param in uv_obj]
+        if k in param_name:
             setattr(uv_obj, k, param_dict[k])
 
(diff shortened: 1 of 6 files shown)


bhazelton:
need to update the documentation with this functionality

pyuvsim/simsetup.py
```diff
@@ -526,7 +526,9 @@ def initialize_uvdata_from_params(obs_params):
         param_dict['object_name'] = '{}_ra{:.4f}_dec{:.4f}'.format(param_dict['sources']['catalog'], src.ra.deg, src.dec.deg)
     uv_obj = UVData()
     for k in param_dict:
-        if hasattr(uv_obj, k):
+        # use the __iter__ function on UVData to get list of UVParameters on UVData
+        param_name = [getattr(uv_obj, param).name for param in uv_obj]
+        if k in param_name:
```
aelanman: I think this block should come after line 549. This way, if the parameter file has `history` or `polarization_array`, etc. those won't be overwritten.

bhazelton: ok, I'll make that change.

bhazelton:
@aelanman I'm re-thinking taking `set_uvws_from_antenna_positions` out of `uvsim.py` -- I'm a little worried that if we start from an existing uvdata object, rather than making one from parameter files, this might cause errors. I could call it in both functions.

Is that possible or is it structured so that the `simsetup.py` code will always be run?

pyuvsim/simsetup.py
bhazelton: although it looks like the history gets overwritten in any case in `init_uvdata_out` in `uvsim.py`

aelanman:
@bhazelton Currently, the options for running off of parameter files and directly from UVData objects are both supported, so there's no guarantee that `initialize_uvdata_from_params` will be run. If running from a file, it will presumably already have uvws, so those won't need to be set.

However, I can imagine a situation where a user simulates off a uvfits file with correct uvws but incorrect antenna positions, expecting pyuvsim to use the uvws, but instead it uses those incorrect antenna positions to make UVWs within the simulator. There should probably be a warning in there if the uvws obtained from antenna positions don't match those already present in the UVData object.

What really should happen at some point is that we write separate `run_file_uvsim` and  `run_param_uvsim` functions that will make their UVData objects and then execute `run_uvsim`, rather than keep the UVData generation calls in the wrappers. It would make for a much cleaner interface.

pyuvsim/simsetup.py
bhazelton: Actually, all of the values that are hardcoded in lines 547-550 are overwritten in `init_uvdata_out`, so moving this won't change anything.

bhazelton:
I think we can't protect users from the situation where they have bad antenna positions, at least not using our current architecture, which fundamentally requires antenna positions.

bhazelton:
@aelanman I added some stuff to the parameter_files.rst to explain the select section. Can you take a look at that file?

pyuvsim/simsetup.py
aelanman: That's true... I guess most development has neglected the situation when someone's simulating off a uvfits file and would want to keep such metadata, but I doubt that will come up often.

I'm starting to think we should deprecate support for simulation off of uvfits files. Each simulation needs more settings than can be given by a uvfits file, and so we'd need to maintain wrappers that provide defaults for those other settings or let the user set them by keyword or something. We have the machinery to make parameter files out of uvfits files anyway.

bhazelton: I'm in favor of heading that direction. Make an issue and we can discuss it with the group.

PR #138, opened 2018-11-06
unpin numpy now that 1.15.4 is out

### LINKED ISSUE #128: numpy is pinned to 1.15.3
lets remove that constraint once numpy/numpy#12248 is resolved

bhazelton:
fixes #128

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -27,7 +27,7 @@ install:
   # Useful for debugging any issues with conda
   - conda info -a
 
-  - conda create -q -n test-environment python=$TRAVIS_PYTHON_VERSION 'numpy <1.15.3' scipy nose pip coverage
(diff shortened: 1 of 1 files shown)


jpober:
Did you update conda?  That is the usually (embarrassing) cause of my inability to update new packages...

bhazelton:
It is updated in .travis.yaml just before the install command, so I don't think that's it. But conda is installing 1.15.4 in Travis for pyuvdata and hera_mc, so I'm a little surprised it isn't here.

bhazelton:
ok, now it's fixed. I streamlined the install process and removed `coverage` from the list of packages to install (we actually use coveralls instead). No idea why those changes fixed it, but I'll take it.

PR #141, opened 2018-11-07
Profiling updates

aelanman:
Support for different modes of profiling batch jobs, and removing a couple unnecessary print statements I'd left in with my last PR.


aelanman:
@bhazelton Sorry, I had one extra change to make (saving time profiling results separately instead of overwriting the same output file). Would you please approve this again?

PR #152, opened 2018-12-01
Update documentation

### LINKED ISSUE #137: Deprecate simulations off of uvfits files.
This mainly a change to the wrappers. We have two wrappers for running simulations: `run_pyuvsim.py`, `run_param_pyuvsim.py`. These both function to accept some input, make a UVData object (and beam_list, beam_dict, and catalog), and then run `run_uvsim` on that object.

Specifying a simulation requires more parameters than can be conveyed in a uvfits file, so I don't think we should continue to support simulations run in this way. Otherwise, we'll have to maintain a separate means of selecting beam lists/paths, source catalogs, approximation parameters, etc. all within a wrapper.

Further, i'd like to remove the `initialize_uvdata_from_params` call from the wrappers and get that contained within `uvsim.py`.

### LINKED ISSUE #151: Update documentation with uvfits deprecation.
Simulations off of uvfits files are deprecated. Update documentation to reflect this.

aelanman:
to reflect deprecation of uvfits sims.

Fixes #151 
Fixes #137

### DIFF
--- a/docs/usage.rst
+++ b/docs/usage.rst
@@ -2,10 +2,11 @@
 Running a simulation
 ====================
 
-``pyuvsim`` takes a UVData object, instrument configuration settings, and a source catalog and effectively "fills" the UVData object with simulated data. The function ``uvsim.run_pyuvsim`` in uvsim.py accepts as input a path to a yaml ``obsparam`` file and writes out the data to ``uvfits`` format. Optionally, it can also skip writing the data out and just return a ``UVData`` object filled with simulated data.
-
+The function ``uvsim.run_uvsim`` in uvsim.py accepts as input a path to a yaml ``obsparam`` file and writes out the data to ``uvfits`` format. Optionally, it can also skip writing the data out and just return a ``UVData`` object filled with simulated data.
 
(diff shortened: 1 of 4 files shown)


docs/usage.rst
```diff
@@ -2,44 +2,27 @@
 Running a simulation
 ====================
 
-``pyuvsim`` takes a UVData object, instrument configuration settings, and a source catalog and effectively "fills" the UVData object with simulated data. The function ``run_pyuvsim`` in uvsim.py
+``pyuvsim`` takes a UVData object, instrument configuration settings, and a source catalog and effectively "fills" the UVData object with simulated data. The function ``uvsim.run_pyuvsim`` in uvsim.py accepts as input a path to a yaml ``obsparam`` file and writes out the data to ``uvfits`` format. Optionally, it can also skip writing the data out and just return a ``UVData`` object filled with simulated data.
```
bhazelton: This text still suggests to me that a UVData object is a standard input. Is that what we want?

aelanman: I wanted to emphasize that the underlying function of pyuvsim is to "fill" a UVData object, but I can rephrase it to avoid giving the impression that uvdata files are supported as input.

PR #153, opened 2018-12-01
Fix memory bug and add coarse horizon cut

### LINKED ISSUE #49: Memory usage
Big job (~3.7 million tasks) exceeded 100 GB memory limit.  Seems to have happened during task list generation.  Obviously an issue, needs investigation.

### LINKED ISSUE #140: Catch warnings in tests
A bunch of warnings have crept in again.

### LINKED ISSUE #147: parameter checking in init_uvdata_out
https://github.com/RadioAstronomySoftwareGroup/pyuvsim/blob/0cd3f1f7f5a097276ec4ac0182205a8e2e906c3e/pyuvsim/uvsim.py#L250

It seems to me that these singular checks should be outside the main "if" statement... you want to check if each is a string in the case that all of them are passed. Otherwise, if not all passed, the main error is fine.

aelanman:
We were overrunning memory limits by accidentally generating `Ntasks` length integer arrays on every rank at the same time. This avoids that.

To better reduce runtime, I've added options to do minimum/maximum flux cuts in catalog reading. There is also a rough calculation in the catalog reader that determines rise and set lsts for sources (with some "buffer" time subtracted/added to them). This will (1) exclude sources that never rise from the catalog altogether and (2) allow for sources definitely below the horizon to be skipped in the task loop. This saves us from doing the AltAz calculation for those. Some below-horizon sources may still make it through this cut, but those are still excluded by the more precise astropy method. The tests should ensure that no above-horizon sources are excluded.

Some other minor adjustments:
- Allows support for writing to uvh5 and miriad formats.
- Fixes a bug with file numbering.
- Improves some test coverage.
- Fixes error with a particular setup parameter case.
- Changes to progsteps to ensure it doesn't print the same progress step more than once.

Fixes #140, Fixes #49 , Fixes #147


CHANGELOG.md
```diff
@@ -3,13 +3,22 @@
 ## [Unreleased]
 
 ### Added
+- Flux cut options with source catalogs
+- Coarse horizon cut for reduced computational load.
+- Selection keys in obsparam files (valid options for UVData.select)
```
bhazelton: thanks for adding the select change!

docs/parameter_files.rst
```diff
@@ -131,6 +131,11 @@ Sources
 
     .. autofunction:: create_mock_catalog
 
+    Flux limits can be made by providing the keywords ``min_flux`` and ``max_flux``. These specify the min/max stokes I flux to choose from the catalog.
+
+    The option ``horizon_buffer`` can be set (in radians) to adjust the tolerance on the coarse horizon cut. After reading in the catalog, ``pyuvsim`` roughly calculates the rise and set times (in local sidereal time, in radians) for each source. If the source never rises, it is excluded from the simulation, and if the source never sets its rise/set times are set to None. This calculation is less accurate than the astropy alt/az calculation used in the main task loop, so a ``buffer'' angle is added to the set lst (and subtracted from the rise lst) to ensure sources aren't accidentally excluded. Tests indicate that a 10 minute buffer is sufficient.
```
bhazelton: I think the quote marks on buffer aren't matched properly.

pyuvsim/simsetup.py
```diff
@@ -44,12 +45,84 @@ def _parse_layout_csv(layout_csv):
                          dtype=dt.dtype)
 
 
-def read_votable_catalog(gleam_votable):
+def _array_to_sourcelist(catalog_table, lst_array=None, time_array=None, latitude_deg=None, horizon_buffer=0.04364, min_flux=None, max_flux=None):
+    """
+    Performs selections on source recarrays from
```
bhazelton: incomplete sentence

pyuvsim/simsetup.py
```diff
@@ -44,12 +45,84 @@ def _parse_layout_csv(layout_csv):
...
+
+
+def read_votable_catalog(gleam_votable, input_uv=None, source_select_kwds={}):
     """
     Creates a list of pyuvsim source objects from a votable catalog.
 
     Args:
         gleam_votable: Path to votable catalog file.
+        input_uv: The UVData object for the simulation (needed for horizon cuts)
+        source_select_kwds: Keywords for source selection (passed along)
```
bhazelton: I'd prefer to define the relevant keywords on this function and then pass them through. Currently, they're only defined on a private function (starts with an underscore), which will not be documented on RTD (it skips private functions).

pyuvsim/simsetup.py
```diff
@@ -44,12 +45,84 @@ def _parse_layout_csv(layout_csv):
...
+
+    Args:
+        catalog_table: recarray of source catalog information.
+        lst_array: For coarse RA horizon cuts, lsts used in the simulation [radians]
+        time_array: Array of (float) julian dates corresponding with lst_array
+        latitude_deg: Latitude of telescope in degrees. Used for declination coarse horizon cut.
+        horizon_buffer: Angle (float, in radians) of buffer for coarse horizon cut. Default is about 10 minutes of sky rotation.
+                        Sources whose calculated altitude is less than -horizon_buffer are excluded by the catalog read.
+                        Caution! The altitude calculation does not account for precession/nutation of the Earth. A decent buffer is needed to
+                        ensure that the horizon cut doesn't exclude sources near the horizon.
```
bhazelton: This worries me a little because how much buffer you need will get larger as a function of the time difference from J2000. Is there a way to be sure that we're always close enough?

pyuvsim/simsetup.py
```diff
@@ -69,32 +142,29 @@ class Found(Exception):
...
-    """
-    with open(filename, 'w+') as fo:
-        fo.write("SOURCE_ID\tRA_J2000 [deg]\tDec_J2000 [deg]\tFlux [Jy]\tFrequency [Hz]\n")
-        for src in catalog:
-            fo.write("{}\t{:f}\t{:f}\t{:0.2f}\t{:0.2f}\n".format(src.name, src.ra.deg, src.dec.deg, src.stokes[0], src.freq.to("Hz").value))
+    return sourcelist
 
 
-def read_text_catalog(catalog_csv):
+def read_text_catalog(catalog_csv, input_uv=None, source_select_kwds={}):
```
bhazelton: The docstring for this function is missing the optional parameters.

pyuvsim/source.py
```diff
@@ -40,6 +40,10 @@ def __init__(self, name, ra, dec, freq, stokes, pos_tol=np.finfo(float).eps):
                 4 element vector giving the source [I, Q, U, V]
             freq: astropy quantity
                 frequency of source catalog value
+            rise_time_inds: int
+                time array indices at which source rises
```
bhazelton: maybe indicate that these are coarse values that need to be conservative?

pyuvsim/simsetup.py
aelanman: Is there some relation between time since J2000 and the accumulated effect of precession/nutation? I got the 10 minutes through trial and error in tests...

pyuvsim/simsetup.py
aelanman: I decided to make the 'array_to_sourcelist` function public. I can still define the options here if you think that'd be better.

pyuvsim/simsetup.py
bhazelton: Yes, it should. That makes me feel better, can you update the docstring to make it clear?

pyuvsim/simsetup.py
bhazelton: I think it is better. I know it involves copy/pasting but it's docs not code and I think people find it easier to read docs that are close to the code.

pyuvsim/source.py
```diff
@@ -40,6 +40,12 @@ def __init__(self, name, ra, dec, freq, stokes, pos_tol=np.finfo(float).eps):
                 4 element vector giving the source [I, Q, U, V]
             freq: astropy quantity
                 frequency of source catalog value
+            rise_lst: (float)
+                Approximate lst (radians) when the source rises. Set by coarse horizon cut in simsetup.
+                Default is None, meaning the source never rises.
+            set_lst: (float)
+                Approximate lst (radians) when the source sets.
+                Default is None, meaning the source never sets.
```
bhazelton: It sounds to me like these may not be really optional parameters -- if you set neither of them what happens?

bhazelton: ok, sounds reasonable.

docs/usage.rst
```diff
@@ -2,44 +2,27 @@
 Running a simulation
 ====================
 
-``pyuvsim`` takes a UVData object, instrument configuration settings, and a source catalog and effectively "fills" the UVData object with simulated data. The function ``run_pyuvsim`` in uvsim.py
+``pyuvsim`` takes a UVData object, instrument configuration settings, and a source catalog and effectively "fills" the UVData object with simulated data. The function ``uvsim.run_pyuvsim`` in uvsim.py accepts as input a path to a yaml ``obsparam`` file and writes out the data to ``uvfits`` format. Optionally, it can also skip writing the data out and just return a ``UVData`` object filled with simulated data.
```
bhazelton: Should this be updated to indicate that miriad files can be written out too? Out of curiosity, why miriad and not uvh5? uvh5 is what HERA is using going forward, it supports both phased and drift data and has good support for partial reading and partial writing.

aelanman: I still need to add a way to specify output format from the wrapper scripts and functions (at the moment, I've just changed the `write_uvdata` function to accept those options). I'll update the docs once that's done.

I went with miriad because I've been borrowing these functions for my healpix sky simulator and need those outputs in miriad. I'll add the uvh5 option too. Is there a common extension those files take?

bhazelton: yes, `.uvh5`

aelanman: Should be addressed now. Users may specify output format through the keyword `output_format` in the `filing` section of the obsparam. Default is to write to uvfits, but uvh5 and miriad are also supported.

aelanman:
This is now seeing the deprecation warnings from astropy and a TypeError from comparing telescope objects (apparently from astropy's unit representation again). This is only coming up in python3.

bhazelton:
I can try to help. The fix I came up with for pyuvdata is here: RadioAstronomySoftwareGroup/pyuvdata#500

It turns out that at least in some cases it needs to be subclassed by python version because astropy 2.0 and 3.1 have incompatible keyword names.

bhazelton:
I just merged in the pyuvdata fix, so I'm going to restart these jobs to see what's coming just from pyuvsim (vs pyuvdata)

bhazelton:
Ok, I think I fixed the residual issue in telescope.py -- apparently in astropy 3.1 earthlocation.value returns something with a numpy.void type which breaks np.allclose. Fixed by explicitly asking for the x, y, z values which have better defined types.

aelanman:
That seems to have done it! Thanks, Bryna.

pyuvsim/uvsim.py
```diff
@@ -355,12 +375,24 @@ def run_uvdata_uvsim(input_uv, beam_list, beam_dict=None, catalog=None, source_l
...
+    Neach_section, extras = divmod(Ntasks, Npus)
+    if rank < extras:
+        length = Neach_section + 1
+        start = rank * (length)
+        end = start + length
+    else:
+        length = Neach_section
+        start = extras * (Neach_section + 1) + (rank - extras) * length
+        end = start + length
+    if six.PY2:
```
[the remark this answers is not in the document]
aelanman: Thanks!

PR #170, opened 2019-01-30
Bessel correction again

aelanman:
Remove power of two from Airy beam (e-field). Also tests that the width of beam in Fourier space is less than the dish diameter in wavelengths.


aelanman:
@bhazelton This should be ready now. I've added a new test that check that the uv-space fwhm of an Airy beam is less than the dish diameter in wavelengths, but I'd like some feedback on it. I think is a good thing to check to make sure our diameter parameter is being used correctly, but I'm not entirely certain of how the FWHM of the uv footprint of a beam should relate to beam width (just that it should be smaller).

pyuvsim/tests/test_analyticbeam.py
```diff
@@ -118,6 +119,44 @@ def test_airy_beam():
...
+    uvals = np.arange(-Npix // 2, Npix // 2) * upix
+    for i, bk in enumerate(beam_kern):
+        thresh = np.abs(bk[Npix // 2, Npix // 2]) * 0.01    # Cutoff at 1% of the center value in Fourier space.
+        points = np.sum(np.abs(bk) >= thresh)
+        area = np.sum(points) * upix**2
+        kern_radius = np.sqrt(area / np.pi)
+
+        # For each frequency, check kernel width is less than dish diameter in wavelengths
+        nt.assert_true(kern_radius < 2 * diameter_m / lams[i])
+        print(kern_radius, 2 * diameter_m / lams[i])
```
bhazelton: not sure we want to keep this print statement. Also, I think we want a test that's a little stronger. I don't just want it to be less, I want it to be close (for some definition of close).

CHANGELOG.md
```diff
@@ -10,6 +10,7 @@
 - Coarse horizon cut for reduced computational load.
 - Selection keys in obsparam files (valid options for UVData.select)
 - New batch scripts for profiling jobs on a SLURM-managed cluster, and plotting the results.
+- Checking the beam kernel width makes sense for Airy beams
```
bhazelton: I'm a little worried that this won't be in the right place if we merge this as is. I think it needs to be rebased and then check that this change is showing up under unreleased, rather than under version 0.2.1

pyuvsim/tests/test_analyticbeam.py
```diff
@@ -8,6 +8,7 @@
 import numpy as np
 import nose.tools as nt
 from scipy.special import j1
+from scipy.interpolate import UnivariateSpline
```
bhazelton: I don't see where you're using this import

pyuvsim/tests/test_analyticbeam.py
aelanman: I'm not sure how best to address this... the widths are close, and seem to scale properly, but are off by a non-negligible amount. For this test, 2*D is about a factor of 1.2 times larger than the beam kernel radius. It's pretty consistently larger by that factor, but I'm not sure where that's coming from.

pyuvsim/tests/test_analyticbeam.py
aelanman: That was for the old version of the test. Thanks for catching it.

CHANGELOG.md
bhazelton: Now that you've rebased, I can tell it's in the wrong place. It needs to be under "unreleased"

pyuvsim/tests/test_analyticbeam.py
bhazelton: Hmm, that is not as small factor as I would expect.

CHANGELOG.md
aelanman: Missed that... should be fixed now.

pyuvsim/tests/test_analyticbeam.py
bhazelton: I'm really worried about this, can we talk about it on tomorrow's telecon?

aelanman: Agreed. I've been fiddling with the test a little, and it looks like the error depends on the dish diameter, but not on the field of view or resolution.

pyuvsim/tests/test_analyticbeam.py
```diff
@@ -118,6 +118,43 @@ def test_airy_beam():
...
+    beam_kern = np.fft.fft2(ebeam, axes=(1, 2))
+    beam_kern = np.fft.fftshift(beam_kern, axes=(1, 2))
+
+    for i, bk in enumerate(beam_kern):
+        thresh = np.max(np.abs(bk)) * 0.005  # Cutoff at half a % of the maximum value in Fourier space.
+        points = np.sum(np.abs(bk) >= thresh)
+        upix = 1 / (2 * np.sin(zmax))   # 2*sin(zmax) = fov extent projected onto the xy plane
+        area = np.sum(points) * upix**2
+        kern_radius = np.sqrt(area / np.pi)
+        print((diameter_m / lams[i]) / kern_radius)
```
bhazelton: do we still want this print statement?

pyuvsim/tests/test_analyticbeam.py
```diff
@@ -118,6 +118,43 @@ def test_airy_beam():
...
+        nt.assert_true(np.isclose(diameter_m / lams[i], kern_radius, rtol=0.5))
```
bhazelton: this is great! Thank you for running this down.

CHANGELOG.md
```diff
@@ -4,6 +4,7 @@
 
 ### Added
 - Support for only simulating one baseline per redundant group
+- Checking the beam kernel width makes sense for Airy beams
```
bhazelton: Can you rebase and check that this line ends up in the right place? I'm a little worried it'll be wrong after the recent version release.

aelanman:
Thanks for your notes, @bhazelton . I've made the changes.

PR #179, opened 2019-02-10
Frequency array definition

aelanman:
Clarifies that the entries in the frequency array refer to channel centers, and fixes the relationships among frequency parameters.

Also applies similar changes to the time array setup.


docs/parameter_files.rst
```diff
@@ -77,8 +84,21 @@ Time and frequency
...
 
     As long as one of the sets from each category above is met by the supplied keywords, the frequency array will be successfully built.
-    Likewise for the time array:
+    You can also just give an explicit ``freq_array``.
+
+
+Time
+^^^^
 
+    The time array is specified similarly. The entries in the ``time_array`` indicate the **beginning of each time step in Julian date**. The ``integration_time`` is the time step size in seconds. The user may also specify ``duration_hours`` or ``duration_days`` to specify the total time covered by all time steps. The following relations among parameters hold::
```
bhazelton: I find this confusing -- it sounds like the time_array values are for the start of the integration, not the center of the integration.

aelanman: That was my intention. Should it be the center of the integration?

bhazelton: I think it should. UVData time_array values correspond to the center of the integration. Of course for pyuvsim, we aren't really integrating, we're just calculating the value at a moment in time. We might want to clarify that in the docs.

aelanman:
With this branch, if I use the following specifications for the frequency channels, i get identical channels to the PRISim file:
`{'bandwidth': 199999488.0, 'channel_width': 32552.0, 'start_freq': 50000256.0}`
`{'Nfreqs': 6144, 'channel_width': 32552.0, 'start_freq': 50000256.0}`
`{'bandwidth': 199999488.0, 'start_freq': 50000256.0, 'Nfreqs': 6144}`

Note that the `start_freq` and `end_freq` refer to the centers of the first and last channel, respectively. This means that if, for instance, we want our bandpass to cover 100 to 200 MHz with a 100 kHz channel width, then the `start_freq` is 100.05 MHz.

I leave it to discussion if this is truly what we want, but this is at least consistent with what PRISim is doing.

docs/parameter_files.rst
aelanman: Okay, that makes sense. I think in terms of the `simsetup.py` code, both standards are written the same. I'll update the docs.

aelanman:
This should be ready for review.

aelanman:
@bhazelton I've added the extra checks I wanted and made some fixes related to recent updates to UVBeam. This should be ready for inspection now.

pyuvsim/simsetup.py
```diff
@@ -745,9 +759,63 @@ def parse_time_params(time_params):
...
+    fdict['channel_width'] = np.diff(freq_array)[0]
+    fdict['Nfreqs'] = freq_array.size
+    fdict['bandwidth'] = fdict['channel_width'] * fdict['Nfreqs']
+    fdict['start_freq'] = freq_array[0]
+    fdict['end_freq'] = freq_array[-1]
+
+    return fdict
+
+
+def time_array_to_params(time_array):
```
bhazelton: Where are these functions used? I only see them used in the tests. The reason I ask, is that while the frequencies have to be evenly spaced for uvfits & miriad, time arrays do not in general have to be equally spaced. So if you're using this function to get params from UVData objects, you shouldn't necessarily assume that they will be evenly spaced.

pyuvsim/simsetup.py
```diff
@@ -715,28 +729,28 @@ def parse_time_params(time_params):
...
         raise ValueError("Either a start or end time must be specified: " + kws_used)
 
     time_arr = np.linspace(time_params['start_time'],
-                           time_params['end_time'],
+                           time_params['end_time'] + inttime_days,
                            time_params['Ntimes'], endpoint=False)
 
     if time_params['Ntimes'] != 1:
-        assert np.allclose(np.diff(time_arr), inttime_days * np.ones(time_params["Ntimes"] - 1), atol=1e-4)   # To nearest second
+        assert np.allclose(np.diff(time_arr), inttime_days * np.ones(time_params["Ntimes"] - 1), atol=dayspersec)   # To nearest second
```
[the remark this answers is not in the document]
bhazelton: Actually, I'm not sure you should assert this at all. Time arrays aren't required to be evenly spaced in UVData objects or in uvfits files.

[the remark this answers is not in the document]
aelanman: Unless the user provides an explicit time_array, simsetup will make an evenly-spaced time array. This is meant to be an extra check that the provided parameters are consistent. I guess I can put this under the `else` block, so it won't be called if the user provides a `time_array`?

pyuvsim/simsetup.py
aelanman: There have been a few times I've had an array of frequencies or times and wished I had a quick way to get the right parameters to make it.

I'll adjust the `time_array_to_params` to check if the parameters are evenly spaced first. If it is, it will return some more compact representation of the time array (e.g., start_time + Ntimes + integration_time). If not, it will just return the full array.

aelanman: (To add -- these functions will be used in the `uvdata_to_config` function)

pyuvsim/simsetup.py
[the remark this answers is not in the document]
bhazelton: that's reasonable.

pyuvsim/simsetup.py
bhazelton: that sounds good.

pyuvsim/simsetup.py
```diff
@@ -607,12 +608,14 @@ def parse_frequency_params(freq_params):
...
     fa, sf, ef, nf, cw, bw = [fk in freq_params for fk in freq_keywords]
     kws_used = ", ".join(freq_params.keys())
+    _freq_params = copy.deepcopy(freq_params)
 
     if fa:
         freq_arr = np.array(freq_params['freq_array'])
         freq_params['Nfreqs'] = freq_arr.size
         if freq_params['Nfreqs'] > 1:
             freq_params['channel_width'] = np.diff(freq_arr)[0]
+            assert np.all(np.diff(freq_arr) == freq_params['channel_width'])
```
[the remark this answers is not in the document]
bhazelton: Looks like this is still just an assert, do you want to leave it that way?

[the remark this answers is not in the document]
aelanman: Whoops... I meant to replace that one too. My last push replaces it.

aelanman:
@bhazelton I've made some further changes to the tests that I've been meaning to do for a while. Since the `parse_frequency_params` (and time) are separate from `initialize_uvdata_from_params`, those functions should be tested separately. We won't need to carry around separate test obsparam files for each parameter configuration.

Rather than fix `check_param_reader` to cover the missing lines, I separated the time/freq array tests into their own functions (including all error cases) and removed some of `check_param_reader`. The new tests are a little more clear about checking every possible valid combination of parameters and confirms that the right errors are raised when the right parameters are missing.

We still need `check_param_reader` to test the telescope configuration and all the `initialize_uvdata_from_params` functions, but it doesn't need to loop over all the separate cases anymore.

PR #181, opened 2019-02-23
Chromatic gaussian beam

### LINKED ISSUE #144: Frequency dependent gaussian beam
Should have option for some simple (e.g. linear) frequency dependence in our analytic beams.

aelanman:
- Enables Gaussian beams to be defined with an antenna diameter, such that its FWHM is equal to that of an Airy disk for the same antenna diameter.
- Begins deprecating Gaussian beams defined with "sigma".

Fixes #144

### DIFF
--- a/pyuvsim/analyticbeam.py
+++ b/pyuvsim/analyticbeam.py
@@ -15,9 +15,6 @@ def diameter_to_sigma(diam, freqs):
     an Airy disk's main lobe for a given diameter.
     """
 
-    if not isinstance(freqs, np.ndarray):
-        freqs = np.array([freqs]).dtype(np.float)
-
     c_ms = 299792458.
     wavelengths = c_ms / freqs
 
--- a/pyuvsim/analyticbeam.py
+++ b/pyuvsim/analyticbeam.py
@@ -95,8 +95,8 @@ class AnalyticBeam(object):
             # standard deviation of sigma is referring to the standard deviation of e-field beam!
             # copy along freq. axis
             if self.diameter is not None:
-                sigmas = diameter_to_sigma(diameter, freq_array)
-                values = np.exp(-(za_array[np.newaxis, ...]**2) / (2 * sigma[:, np.newaxis]**2))
+                sigmas = diameter_to_sigma(self.diameter, freq_array)
+                values = np.exp(-(za_array[np.newaxis, ...]**2) / (2 * sigmas[:, np.newaxis]**2))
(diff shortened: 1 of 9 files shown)


pyuvsim/analyticbeam.py
```diff
@@ -5,9 +5,26 @@
 from __future__ import absolute_import, division, print_function
 
 import numpy as np
+import warnings
 from scipy.special import j1
 
 
+def diameter_to_sigma(diam, freqs):
```
anguta: I would make sure to explicitly state that "sigma" here is in units of the polar angle rather than units of "angle cosines" which I believe it is often defined by 

https://en.wikipedia.org/wiki/Airy_disk#Approximation_using_a_Gaussian_profile

You may even want to provide an option for both.

pyuvsim/analyticbeam.py
```diff
@@ -64,22 +85,26 @@ def interp(self, az_array, za_array, freq_array, reuse_spline=None):
...
             interp_basis_vector = None
         elif self.type == 'gaussian':
-            if self.sigma is None:
-                raise ValueError("Sigma needed for gaussian beam -- units: radians")
+            if (self.diameter is None) and (self.sigma is None):
+                raise ValueError("Dish diameter needed for gaussian beam -- units: meters")
             interp_data = np.zeros((2, 1, 2, freq_array.size, az_array.size), dtype=np.float)
             # gaussian beam only depends on Zenith Angle (symmetric is azimuth)
-            # standard deviation of sigma is refereing to the standard deviation of e-field beam!
-            values = np.exp(-(za_array**2) / (2 * self.sigma**2))
```
anguta: The Gaussian here is in units of zenith-angle rather than angle cosines. Since angle cosines are often employed in the Gaussian approximation of an airy beam I would make sure to be explicit about the choice of angle units and consider adding an option for a gaussian of angle cosines.

aelanman: All of our primary beam models are evaluated at angular positions, no angle cosines, as indicated by the labeling of `za_array` and `az_array`. Throughout pyuvsim, there's no point where a direction cosine is used, and since we're not taking any flat sky approximations I can't think of a case where we'd need one.

I'll add a bit to the documentation to confirm that the  that the gaussian beam "sigma" parameter is in zenith angle. I will also make it more clear in the docstring for `diameter_to_sigma`, since sigma is

PR #192, opened 2019-04-03
Joss revisions

aelanman:
Addressing the JOSS reviewer's comments. The biggest change is adding a comparison with other simulators to the documentation, as discussed in the datasim channel.

I want to keep this open until I'm sure that all the comments have been addressed.


aelanman:
Once #179 is merged, I can rebase this and the tests should pass.

docs/comparison.rst
```diff
@@ -0,0 +1,83 @@
+
+Comparison with other simulators
+================================
+
+pyuvsim is designed to calculate visibilities from point sources and a given primary beam model. It does this through direct calculation of the full-polarization Radio Interferometer Measurement Equation (RIME), fully parallelized across a set of MPI processes. Its development has progressed with the goal of providing a tool that can be used reliably and flexibly, but most of all it has focused on avoiding effects that confound foreground modeling efforts in 21cm cosmology.
+
+As such, it has relatively poor performance compared with several other simulators. Deciding which tool to use will ultimately depend on what factors are important to the user. We recommend pyuvsim as a reference tool to verify the results of other simulators, and for producing limited datasets for comparison with observations (particularly on long baselines). Users will need access to a lot of memory and a large number of processor cores to take advantage of pyuvsim's accuracy and precision, ideally on a dedicated high-performance computing cluster.
```
bhazelton: maybe "poor performance" -> "poor speed" or "slow"

docs/comparison.rst
```diff
@@ -0,0 +1,83 @@
...
+2. Well-tested through years of use.
+3. Support for a range of sky models including point sources, diffuse emission, and spectral cubes.
+4. Support for tracking as well as transit telescopes.
+5. Parallelized over baselines, frequencies, or sky model/catalog.
+6. Has been tested using imaging with CASA.
+
+Disadvantages:
+
+1. MPI overhead is large if time axis in simulation is long.
+2. Supports only one polarization at a time.
```
bhazelton: Also can only support real power beams, so not cross-polarizations (e.g. PRISim supports 'XX', 'YY' but not 'XY', 'YX')

docs/comparison.rst
```diff
@@ -0,0 +1,83 @@
...
+3. Does not support non-identical antenna patterns.
+
+Fast Holographic Deconvolution (FHD)
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+FHD is a data calibration and modeling framework written in IDL by researchers at the University of Washington for analyzing data from the Murchison Widefield Array (MWA) in Western Australia. It calculates sky models by convolving a high-resolution psf with the exact DFT of point sources. This is approach is more efficient when working with wide primary beams, which will be narrower in Fourier space and thus limit the size of the convolution to be performed.
```
bhazelton: I think the last sentence is true but it's not a very large effect, so highlighting it so prominently feels a little weird to me.

docs/comparison.rst
```diff
@@ -0,0 +1,83 @@
...
+5. Uses IDL's native parallelization efficiently.
+
+Disadvantages:
+
+1. Written in the proprietary IDL language.
+2. Designed for the MWA, and has limited support for other instruments.
+3. The discrete convolution in Fourier Space can introduce aliasing and ringing artifacts at a level relevant to 21cm cosmology.
+4. Measurement equation formalism relies on the flat-sky approximation, limiting curved sky effects.
+5. Primary beam motion on the sky is limited to the "snapshot" size, not the time step size.
+6. Not parallelized over frequency.
```
bhazelton: This last point is a bit confusing. It's true that visibilities are calculated separately by frequency, but they all use a single UV plane, so there's only one DFT of all the sources.

docs/comparison.rst
```diff
@@ -0,0 +1,83 @@
...
+3. Support for diffuse foreground models.
+4. Support for discrete pointings.
+5. Uses IDL's native parallelization efficiently.
+
+Disadvantages:
+
+1. Written in the proprietary IDL language.
+2. Designed for the MWA, and has limited support for other instruments.
+3. The discrete convolution in Fourier Space can introduce aliasing and ringing artifacts at a level relevant to 21cm cosmology.
+4. Measurement equation formalism relies on the flat-sky approximation, limiting curved sky effects.
```
bhazelton: I need to talk to Ian about this, I know there are some curved sky effects handled in FHD.

docs/comparison.rst
```diff
@@ -0,0 +1,83 @@
...
+
+1. Uses compiled C code internally, which is faster.
+2. Established in the field already and very well-documented.
+3. OpenMP utilizes shared memory on a single node if the calculation an be decomposed efficiently. MPI for all other parallel processing needs.
+4. Support for a source component lists and FITS image source models.
+
+Disadvantages:
+
+1. Limited support for user-defined primary beam models.
+2. Lacks full-sky coverage.
```
bhazelton: Reword this. There are 2 potential issues:
1) the measurement equation it uses is almost certainly not correct for wide-field sources (Nithya will check on wording here)
2)The internal uvw rotation (for phasing) is known to be incorrect, this affects the coherence of sources far from the phase center.

docs/comparison.rst
```diff
@@ -0,0 +1,83 @@
...
+3. Point sources are gridded to pixel locations, losing precision. This pixel-scale imprecision can introduce point source subtraction errors that are significant to 21cm cosmology experiments [CTROTT2012]_.
+4. Full direction-dependent Jones matrices are supported, but simulating with them requires that the beam times sky model calculation be carried out in separate softwa [JAGANNATHAN17]_.
```
bhazelton: softwa -> software

bhazelton:
I need to update the setup.py to work better.

docs/comparison.rst
miguelfmorales: Yeah, and it is not actually accurate compared to other uv-integrators. The sampling pitch of the uv plane _relative_ to the antenna size is what determines the speed, and these are usually chosen at a fixed ratio. So for example a VLA antenna at 40 GHz is very large in wavelengths, but if you pick the uv sampling to go out 2 or 3 sidelobes in the antenna pattern the speed is the same as for the MWA antenna. 

I might also adjust "by researchers at the University of Washington for analyzing data from the Murchison Widefield Array (MWA) in Western Australia." to "by researchers at the University of Washington, originally developed for analyzing EoR data from the Murchison Widefield Array (MWA) in Western Australia."

docs/comparison.rst
miguelfmorales: Yeah, 4 is mostly wrong and I'd just delete. We wrote FHD _because_ we needed the curved sky effects. This is literally why we started writing it. So FHD has all the curved sky effects in, with the small caveat that it cannot model the sky more than 90° away from the phase center (uv integration approaches fall down in this limit). 

2 is also strange. It certainly has been tested almost exclusively on MWA data, but we have made images of PAPER and HERA data and simulated all sorts of other arrays. Importing beams has been a sticking point, but this should be resolved in a matter of weeks.

miguelfmorales: Trying to think of disadvantages to add. 

One is that high resolution performance is largely untested (but this might be doubly true of the others). 

Another is that it memory scales with the size of the uv-plane. This helps it be very fast for compact 21 cm cosmology arrays, but for higher resolution arrays with sparse uv coverage the memory scaling can be poor.

docs/comparison.rst
miguelfmorales: Yeah, 6 is nuanced. Part of the speed is that it does not redo uv plane calculations of smooth spectrum sources. Is this infinitely parallel (did all the calculations at once!) or not parallel at all?

setup.py
```diff
@@ -30,8 +30,10 @@
     'scripts': glob.glob('scripts/*'),
     'version': version.version,
     'include_package_data': True,
-    'install_requires': ['numpy>=1.10', 'astropy>=1.2', 'scipy', 'mpi4py',
-                         'pyyaml', 'six>=1.10', 'pyuvdata'],
+    'test_suite': 'nose.collector'
+    'test_requires': ['nose', 'h5py'],
+    'install_requires': ['numpy>=1.14', 'astropy>=2.0', 'scipy', 'mpi4py>1.3.1',
+                         'pyyaml', 'six>=1.10', 'pyuvdata>=1.3'],
```
bhazelton: I think the mpi4py version requirement should be higher than this, but I don't know what it should be.
I think they pyuvdata version requirement may also need to be higher, but I'm not sure what it should be.

bhazelton:
You might add a note to the reviewer that we are working on a conda package to help with installation.

docs/comparison.rst
bhazelton: I think this means it cannot do accurately simulate polarized sources.

nithyanandan: I'd state it as "Limited to accurately simulating the scalar (total intensity or diagonal components of) radio interferometer measurement equation if either the polarized sky emission and/or the off-diagonal terms in beam pattern are/is negligible." Or something along these lines. I think that's a fair and objective statement.

docs/comparison.rst
nithyanandan: I spoke recently to some CASA developers. They claim they can achieve full sky coverage if the model and the beam (or model times custom beam) covers the full sky. They also claim to achieve it accurately without making flat-sky or narrow-field approximations by using the brute force DFT equation. So stating "lacks full-sky coverage" is not accurate. I don't know about the phasing issues though, which sounds like a different issue. If we want to insist on that point of phasing accuracy, perhaps we need to quote some reference or outstanding ticket that acknowledges that problem.

bhazelton: @dannyjacobs can you get us a ticket number?

docs/comparison.rst
bhazelton: Ok. I do think that clarifying that 'XX' and 'YY' are supported but 'XY' and 'YX' aren't in some sort of parenthetical would help me understand what your sentence means.

nithyanandan: Let me try a longer explanation. What I am referring to is the standard leakage from polarized emission to total intensity through off-diagonal terms in the beam and a polarized sky, through a matrix product of the Mueller beam and the Stokes Intensity. For this leakage to be significant, significant polarized emission and significant off-diagonal beam terms are both essential. If either one of them is negligible, then the leakage is negligible too. I was just stating that PRISim works accurately for limited applications where the leakage is negligible, in other words when the matrix product can be effectively scalarized because the product of off-diagonal beam times polarized emission is negligible.

docs/comparison.rst
aelanman: For 6, I was thinking of the loop over frequency done by `vis_simulate.pro` if `model_uvf_arr` has a frequency axis. This is the case for simulating EoR-like signals, and has been a big slowdown for my work with FHD in the past. I'm not sure how much support there is for signal spectral modeling (vs. foregrounds).

docs/comparison.rst
aelanman: I'll remove (4). I was thinking mainly about how the visibility equation can be written as a 2d Fourier transform only in a flat sky approximation, but of course FHD is sophisticated enough to handle the curved sky effects.

setup.py
aelanman: `mpi4py` is a tricky requirement. The conda package installs a particular version of MPI with it, and wraps those functions, but it can be built against any MPI installation. That's what's currently done in Travis in order to support MPI-3 functions. I'd like to stick to MPI-3 so we can use shared memory in the future, but I'm not sure what the minimum mpi4py requirement is to support that (I think it's had wrappers for MPI-3 functions for some time...). At the very least, we need a version of mpi4py that's Python-3 compatible.

The newest pyuvdata function I can think of that's required is Nick's new frequency interpolation for healpix beams. So whatever version was in place on March 30th should be the minimum requirement.

aelanman:
Thanks everyone for their comments and corrections. I've been traveling this weekend and so I'm going to be a bit slower to respond.

docs/comparison.rst
miguelfmorales: Yeah, FHD does this simply by working in ortho-slant image coordinates+w-projection snapshots, which is a fairly standard fix for this issue.

docs/comparison.rst
miguelfmorales: Yeah, this is why it is subtle. FHD is very slow for EoR signals, but it was designed to simulate smooth spectrum foregrounds and at that task it is fast.

Just for dissimenating information, in typical useage all the foregrounds are assumed to have the same spectral index over the simulated band. But the design allows for the foregrounds to be Taylor expanded in spectral shape, so sources with different spectral indices have a different linear term in the Taylor expansion and sources with curvature pick up higher terms. The simulation speed scales with the number of terms in the Taylor expansion. So even if you have thousands of frequency channels, a third order spectral foreground model is only three times slower than a single frequency channel (holding visibility # constant). 

This of course falls apart for the EoR, which is the worst case scenario for this kind of simulation approach. In this limit it falls back to the same approach as pyuvsim (different model for each frequency).

aelanman: Would this be a fair summary of this discussion, for inclusion on the list?

>FHD is optimized for smooth-spectrum sources but is not designed to handle complex spectral structure. 

I can probably split this sentence between "advantages" and "disadvantages".

docs/comparison.rst
aelanman: @nithyanandan When you say "brute force DFT", does this mean they're not gridding the source model anymore in that case?

aelanman: (never mind, I just saw your previous message in the datasim channel)

setup.py
bhazelton: numpy should be >=1.15

docs/comparison.rst
bhazelton: I just took a crack at re-writing this, does the following work for you @nithyanandan?

Limited to accurately simulating the total intensity (stokes I) radio interferometer measurement equation, which is accurate if the circularly polarized sky emission is negligible and
either the linearly polarized sky emission is negligible or the off-diagonal terms
in beam pattern (e.g. XY or YX for linear polarization) are negligible.

bhazelton: Actually that's not correct either. If there's true stokes U on the sky, you can't see it with any combination of 'XX' and 'YY'. The only polarized emission you can see with those is Stokes Q.

bhazelton: So my version reads:

Limited to accurately simulating the total intensity (stokes I) radio interferometer measurement equation,
which is accurate the polarized sky emission is negligible and the off-diagonal terms
in beam pattern (e.g. XY or YX for linear polarization) are negligible.

requirements.txt
```diff
@@ -0,0 +1,7 @@
+numpy>=1.15
+six>=1.11.0
+setuptools>=38.4.0
+scipy>=1.0.1
+PyYAML>=5.1
+mpi4py>=3.0.0
+astropy>=2.0.6
```
bhazelton: pyuvdata should be on here too. This list should exactly match what's in the setup.py install-requires. There should not be differences between the versions listed here and there.

aelanman: Removing `pyuvdata` from here made it possible to install using `pip install -r requirements.txt`. Keeping it in, the installation fails because it can't find numpy.

This is to make it possible to install in a completely bare-bones environment with only the standard libraries. This is likely overkill... and I'm also still not clear on what we standard practice is for requirements vs. install-requires. There's that distinction between "package" vs. "application" that I don't quite understand.

bhazelton: ok. But I still don't understand the version numbers here being different than the ones in setup.py

aelanman: Good point. It should be fixed now.

docs/comparison.rst
nithyanandan: Sounds good.

aelanman:
@bhazelton Is there anything else this needs before we can merge it?

bhazelton:
I made a couple very small edits, it's ready to go from my perspective.

aelanman:
Thanks @bhazelton ! The phrasing did end up a little weird from me. Do @nithyanandan, @dannyjacobs , @jpober , @jaguirre  or @mkolopanis have any further comments on this PR?

docs/comparison.rst
```diff
@@ -0,0 +1,93 @@
+
+Comparison with other simulators
```
jpober: When viewed in rst format on github, Disadvantages for the various simulators are not well formatted into enumerated lists

README.md
```diff
@@ -5,20 +5,37 @@
...
+pip install .
+```
+* `pyuvsim` is mainly intended to run on clusters running the linux operating system.
+--
+**Note**
+The `mpi4py` module is installed as a wrapper around an existing installation of MPI. The easiest way to install it is to use anaconda, which will install a compatible version of MPI and configure mpi4py to use it:
+```
+conda install -c conda-forge mpi4py
+```
+--
```
jpober: Viewed through github, the string "--" is getting printed.

docs/comparison.rst
bhazelton: This is an issue on RTD too

bhazelton: I still see a problem even with the new updates. I seems to only be a problem in the disadvantages list.

aelanman:
Fixing the "Note" block syntax in markdown somehow hid the block in the index.rst file that pypandocs generated. It should be ready now.

PR #193, opened 2019-04-06
Modularize UVData Initialization in simsetup.py

nkern:
This PR makes `UVData` initialization functions more modular, enabling these functions to be more easily used by packages outside of `pyuvsim` that may not make the same assumptions about how certain arguments are passed.

Principally, this adds `simsetup.setup_uvdata`, which creates an empty UVData object from its documented kwargs. This function can be run without any notion of a telesocpe_config or array_layout YAML file. The API of `simsetup.initialize_uvdata_from_params` hasn't been changed, but now wraps around `setup_uvdata` to achieve initialization. Tests are added.

To achieve this, added functionality was added to`parse_telescope_params`, which allows for `array_layout` to be either a `.csv` file or just a plain dictionary, as well as allowing for no `telescope_config_name` to be passed, in which case it looks for `telescope_location` and `telescope_name` within `tele_params` itself and generates relevant antenna information and returns without searching for beam info. This is useful b/c there are scenarios where other simulators (such as `healvis`) don't make the same assumptions about how beam information is tied to each antenna, and therefore we should be able to setup array info without explicitly required beam info as well.

Also minor bug fixes with inplace dictionary editing in some of the `parse_` functions.

@aelanman the only API change is that `initialize_uvdata_from_params` now returns a fully-constructed `UVData` object, in the sense that all `Nblts`  arrays are created. This was done because certain routines that used its output were selecting on `bls` axis, which requires `ant_1_array` & `ant_2_array` to be populated, but b/c it is an `Nblts` axis it wasn't. Let me know if this capability of having incomplete `UVData` objects returned by this function is desired, it is as simple as turning Line 949 to `fill_blts=False`.


nkern:
@bhazelton is there a standardization in `UVData` that says `ant_1_array` should be either `[0, 0, 0, 0, 1, 1, 1, 2, 2, 3]` or `[0, 1, 2, 3, 1, 2, 3, 2, 3, 3]`? Or is it arbitrary and up to the user?

pyuvsim/simsetup.py
```diff
@@ -487,12 +496,15 @@ def initialize_catalog_from_params(obs_params, input_uv=None):
     return np.array(catalog), source_list_name
 
 
-def parse_telescope_params(tele_params, config_path):
+def parse_telescope_params(tele_params, config_path=''):
     """
     Parse the "telescope" section of obsparam.
```
aelanman: The main thing that concerns me about this function is that the `telescope_config` file also specifies the beam dictionary, which is used to assign beam models to antennas. I can't see where else (absent a telescope config file) the beam models are loaded and assigned.

pyuvsim/tests/test_simsetup.py
```diff
@@ -362,6 +381,15 @@ def test_freq_time_params():
     ttest = pyuvsim.simsetup.parse_time_params(time_dict)
     nt.assert_true(np.allclose(ttest['time_array'], times))
 
+    # check Ntimes = 1 and Nfreqs = 1 case
+    times = np.linspace(2458570, 2458570.5, 1)
+    tdict = pyuvsim.simsetup.time_array_to_params(times)
```
aelanman: Why is this in the `test_freq_parser` test? Shouldn't it be in `test_time_parser`?

pyuvsim/simsetup.py
```diff
@@ -879,70 +939,45 @@ def initialize_uvdata_from_params(obs_params):
...
             if isinstance(bls, six.string_types):
                 # If read from file, this should be a string.
                 bls = ast.literal_eval(bls)
                 select_params['bls'] = bls
             if any([len(item) == 3 for item in bls]):
-                raise ValueError('Only length 2 tuples allowed in bls: can not '
-                                 'down select on polarizations -- pyuvsim '
-                                 'computes all polarizations')
+                raise ValueError('Only length 2 tuples allowed in bls')
+            # don't need it because it was already
```
aelanman: used?

pyuvsim/simsetup.py
```diff
@@ -502,39 +514,94 @@ def parse_telescope_params(tele_params, config_path):
...
+            telescope_location = list(telescope_location_latlonalt)
+            telescope_location[0] *= np.pi / 180.
+            telescope_location[1] *= np.pi / 180.   # Convert to radians
+            tele_params['telescope_location'] = uvutils.XYZ_from_LatLonAlt(*telescope_location)
+            telescope_name = telconfig['telescope_name']
+
+    else:
+        # if not provided, get bare-minumum keys from tele_params
+        if 'telescope_location' not in tele_params:
+            raise KeyError("If telescope_config_name not provided in tele_params, must provide telescope_location")
```
aelanman: Since `tele_params` is an internally-used variable name, this should probably be rephrased to something clearer to the user.

pyuvsim/simsetup.py
```diff
@@ -39,8 +39,17 @@ def _parse_layout_csv(layout_csv):
         str_format_code = 'a'
     else:
         str_format_code = 'U'
-    dt = np.format_parser([str_format_code + '10', 'i4', 'i4', 'f8', 'f8', 'f8'],
-                          ['name', 'number', 'beamid', 'e', 'n', 'u'], header)
+
+    # get data types for each column
+    dtypes = {"name": str_format_code + '10', 'number': 'i4', 'beamid': 'i4', 'e': 'f8', 'n': 'f8', 'u': 'f8'}
```
aelanman: We should probably mention somewhere that the antenna names in layout files are limited to 10 characters.

pyuvsim/simsetup.py
nkern: So if there isn't a telescope config then it returns empty lists and dicts for the beam info. The point there being that some users may want the antenna array functionality of this function without wanting to have to specify beam models (indeed like healvis for example!) In which case you can feed without a telescope config and it will just do the antenna array calculations. The empty beam info are still returned, though, in order to keep the output collection consistent.

aelanman: I think the beam setup should be included with telescope setup, personally, but if we're going to offer separate functions for layout+metadata and beams, then the option to get the beam list/dictionary should be removed from this function.

nkern: Hmm, I think we may be confused. There is only one function to do telescope and beam setup: its `parse_telescope_params`. What I've done is enabled it to _only_ do the telescope setup if the beam information is not present, instead of failing, such that `healvis` etc. can use this function to do what they need to do without it failing. Does that make sense?

bhazelton: I really think we want to keep the required beam functionality -- healvis requires a beam too, yes? If the issue is that healvis can only accept one beam, then there should be an error raised in healvis if multiple beams are specified.

nkern: Hmm, well if the entire philosophy here is to make these functions more usable by packages outside of `pyuvsim`, then we will need to make these things more flexible, otherwise it will simply be much easier for other packages to roll their own simsetup functions (as we did in `healvis` to begin with because it was simply easier than trying to force `pyuvsim`'s adopted routines into `healvis`'s goals).

I think this is a fairly reasonable extension, keeping all functionality that `pyuvsim` requires but enabling it to be used in ways that other simulators might find useful, such as not needing to specify beam parameters when setting up a telescope array (which I think is a really reasonable thing to want to do!). So I guess I say, if we want `pyuvsim` routines to be more acccessible by other simulators, then no, we don't want to keep the required beam functionality. If you want `pyuvsim` specifically to retain that functionality, we can always do those checks in the wrapper functions that call `parse_telescope_params`...

pyuvsim/tests/test_simsetup.py
```diff
@@ -433,11 +473,76 @@ def test_param_select_redundant():
...
+                                        telescope_name="HERA", Nfreqs=10, start_freq=1e8, bandwidth=1e8, Ntimes=60,
+                                        integration_time=100.0, start_time=2458101.0, polarizations=['xx'], no_autos=True,
+                                        fill_blts=False, run_check=True)
+    attrs = ['data_array', 'nsample_array', 'flag_array', 'integration_time', 'ant_1_array', 'ant_2_array']
+    for attr in attrs:
+        nt.assert_true(getattr(uvd, attr) is None)
+
+    # test feeding array layout as dictionary
+    antpos, ants = uvd.get_ENU_antpos()
+    antpos_d = dict(zip(ants, antpos))
```
aelanman: There shouldn't be two forms for the `antenna_layout` can be passed to `parse_telescope_params`. When read from a csv file, it becomes a record array with keys for antenna name, number, beam_ID, E, N, and U. When passed as a dictionary, the only information carried are antenna numbers and positions. Maybe `antenna_layout_csv` is a misleading name since it specifies more than just the layout.

pyuvsim/simsetup.py
aelanman: Thinking about this more, the problem seems to be that `parse_telescope_params` is serving two roles. The first is to interpret the standard obsparam YAML file keywords to extract required metadata for a UVData object, which of course doesn't require any beam information. The second is to set up all required telescope-related information for a simulation, which should include all beam models and their assignments to antennas. The way we've defined our setup yaml files necessarily intertwines these roles. The reason we defined these files, and deprecated simulating off of files, is that UVData doesn't provide sufficient information for a simulation, and we wanted to ensure that our setup standard would completely define a simulation.

I can accept this solution as a way of serving these two roles simultaneously, though I'm still a bit confused about the usage case for running this without paths to the config files in the dictionary. When run with no paths to external files, all this function appears to do is transform ENU antpos to ECEF. When run with reference to a layout file but not telescope_config, it has all the information needed to assign beams to antennas but no beams, so we're ignoring the "BeamIDs" column altogether. Overall, we're passing in a dictionary with all the required parameters for a UVData object and then getting back a slightly changed dictionary with the same information.

pyuvsim/simsetup.py
```diff
@@ -879,70 +939,45 @@ def initialize_uvdata_from_params(obs_params):
...
 
-    # Now make a UVData object with these settings built in.
-    # The syntax below allows for other valid uvdata keywords to be passed
-    #  without explicitly setting them here.
+    # setup object
+    if 'select' in param_dict:
+        uvparam_dict.update(param_dict['select'])
+    if 'object_name' in param_dict:
+        uvparam_dict['object_name'] = param_dict['object_name']
+    uv_obj = setup_uvdata(fill_blts=True, **uvparam_dict)
```
aelanman: I'm not sure I like that `initialize_uvdata_object` calls `parse_telescope_params` and then calls `setup_uvdata`, which itself calls `parse_telescope_params`.

aelanman:
To elaborate on my comments: Our main goal here is to set standards such that simulated products can be compared. Since the output data file doesn't specify all details of the simulation made it, the setup files should serve that purpose. The most general simulator needs to be able to specify a position and beam model for each antenna in an array, and so our yaml specification does that. If a simulator is _not_ trying to use the yaml file standard (like `hera_sim`), it probably shouldn't use these functions since it would have results that cannot be compared with a `pyuvsim` simulation. Since `healvis` is designed to be comparable to a special case of `pyuvsim` (identical power beams), it should be run off the yaml file standard. The issue then is that `healvis` has its own analytic beam models. If `pyuvsim` is supposed to interpret beam information as well as layout and metadata, then it needs to be able to setup healvis-compatible beam models to be useful.

Meeting this standard will require the following:
1. The analytic beam models in `healvis` will need to be moved to `pyuvsim`. There is already a lot of repetition there, they're just different in what they return (power beam values vs. e-field components). I have a PR open now which makes some changes to `pyuvsim`'s analytic beam model to support power beams and power-law-width gaussian beams.
2. The FoV keyword used in `healvis` can be used in pyuvsim to set the horizon cutoff.
3. The `skymodel` keywords used by healvis should be moved into the `sources` section of the obsparam file, and if HEALPix-related keywords are used for a pyuvsim run it should raise a `NotImplementedError`. (When the vectorized sources changes are merged, it's likely we'll start renaming "sources" to "skymodel" anyway).

These changes should make healvis / pyuvsim yaml files completely compatible, and make it possible to compare simulations easily.

Further, I like the idea of making a separate `telescope_config` yaml file unnecessary. This should be an optional format, but the standard entries of the telescope_config yaml must all be present in the `telescope` section, including `telescope_name`, `beam_paths`, and `telescope_location`.

steven-murray:
If I can throw in my 2 cents, I think @aelanman is right. The purpose of pyuvsim as I understand it is to provide a complete description of a simulation (as well as a simulator). If this _is_ the complete description, then I think it should be entirely concrete -- it should not be flexible in the sense that you should not be able to have valid configurations that are structurally different (eg. single-beam and beam-per-antenna). 

If this is so, then there are two options for more "simplistic" simulators: 1) they provide default conversions of their setup to a pyuvsim-based setup (i.e. just make a list of repeating beams to pass to the construction function). 2) Provide common conversions such as these within pyuvsim itself. In this case it would not be a stretch to envision that such conversions may act by default on input arguments to the construction function, as @nkern has suggested, *but* the output from this function should always be completely conformant to the pyuvsim "standard".

aelanman:
If you rebase this it should pass tests again. `pyuvdata`'s warning checker now requires that you provide the message string of the warning. This has been fixed in the tests in more recent commits.

PR #194, opened 2019-04-13
Support spectral gaussian and power beams

aelanman:
This is a first step toward moving functions in `healvis` back to `pyuvsim`. `healvis` simulations use analytic power beams, which are not currently available in `pyuvsim`, and has the option to use chromatic gaussians whose widths change with frequency according to a user-specified power law.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,8 @@
 ## [Unreleased]
 
 ### Added
+- Option to convert analytic beams from efield to pstokes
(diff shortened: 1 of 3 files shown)


aelanman:
@bhazelton I don't know why the coveralls comment hasn't updated, but the tests are all passing.

bhazelton:
It looks like line 84 in analyticbeam.py isn't covered.

aelanman:
Looks like I wasn't calling `assert_raises` correctly. The missing line should be covered now.

PR #197, opened 2019-05-07
Final joss revision

aelanman:
Addressing the remaining comments from the JOSS reviewer. These are pretty minor changes. My main holdup was trying to replicate his error with a Docker container... I was able to install the package and its dependencies and run all tests within the Docker environment running Ubuntu 18.04.

There is a remaining issue with installation in an empty system. Running `pip install -r requirements.txt` will fail because `pyuvdata` cannot be built without some other packages listed, and pip tries to _build_ all packages before installing them. This may be a problem more with pyuvdata than pyuvsim, and isn't essential to getting this approved, but it's worth talking about.

### DIFF
--- a/README.md
+++ b/README.md
@@ -29,9 +29,10 @@ pyuvsim's two primary goals are interferometer simulation accuracy at the level 
 * To install the development version: Clone the repository using `git clone https://github.com/RadioAstronomySoftwareGroup/pyuvsim`, navigate into the pyuvsim directory and run
 ```
 pip install -r requirements.txt
+pip install pyuvdata
 pip install .
 ```
-* `pyuvsim` is mainly intended to run on clusters running the linux operating system.
+* `pyuvsim` is intended to run on clusters running the linux operating system.
 
 
 **Note**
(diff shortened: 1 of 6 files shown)


aelanman:
@bhazelton This should be ready for your review. I think I've met all the reviewer's requests. If this can get merged, I'll reply to the thread and get the paper finished.

README.md
```diff
@@ -41,6 +45,7 @@ conda install -c conda-forge mpi4py
 ## Dependencies
 * `numpy`, `astropy`, `scipy`, `mpi4py`, `pyyaml`, `six`, `pyuvdata`
 * optionally `line_profiler` if you want to do profiling (support for profiling is built in)
+* `h5py` is required to support the new `uvh5` data format.
```
bhazelton: is it an optional or required dependency?

aelanman: Required for that feature of pyuvdata, but not for pyuvsim.

bhazelton: I think you should spell that out a bit more. Maybe something like:

optionally `h5py` to use pyuvdata's `uvh5` file format

aelanman:
Thanks @bhazelton !

PR #198, opened 2019-05-08
Vector sources

### LINKED ISSUE #196: Test visibility calculation with analytic beams
We don't currently have a test that calculates a visibility analytically and compares with a simulated visibility using an analytic beam. I recently tried running a 1Jy zenith point source test with an analytic beam and got 1Jy in both XX and YY (should be 0.5). This was because the off-diagonal terms in the analytic Jones matrix were set to the same values as the diagonal terms.

There could probably be more discussion of what analytic E-field beams should be. Gaussian beams for example seem to be best defined as power beams. If we start adding more sophisticated analytic beam models (e.g., dipole beams), the e-field form may be better defined.

aelanman:
Major algorithmic changes to improve performance by making coordinate transformations and beam interpolation _vectorized_ operations.

- Replaces the `Source` class with `SkyModel`, an object representing a set of point source components.
- The separate `Source.alt_az_calc` method is replaced with `SkyModelupdate_positions`, which calculates alt/az as well as `lmn` positions for all source components.
- Throughout setup, the source list is made and passed as a numpy record array. This makes it easier to broadcast and place in shared memory windows. Data are put into SkyModel objects when they need to be used for coordinate transformations.
- Implements a shared memory option for numpy arrays. This is used to ensure that the catalog is not copied more than `Nnodes` times:
https://github.com/RadioAstronomySoftwareGroup/pyuvsim/blob/af54dec9c65ed3ac31e7d1505a21c030745dffbe/pyuvsim/uvsim.py#L526
- Splits over the "instrument task" axes (time/frequency/baseline) in most cases. It will split over sources instead if `Nbltf < Npus` and `Nsrcs > Npus`. This is done to leverage the speed of vectorized calculations over the potential memory issue of copying the source catalog, but there is a method to avoid overrunning available memory.
- If it is predicted that the SkyModel objects will take up too much memory when generated on each rank, it will loop over subsets of the source array:
https://github.com/RadioAstronomySoftwareGroup/pyuvsim/blob/af54dec9c65ed3ac31e7d1505a21c030745dffbe/pyuvsim/uvsim.py#L211-L217
https://github.com/RadioAstronomySoftwareGroup/pyuvsim/blob/af54dec9c65ed3ac31e7d1505a21c030745dffbe/pyuvsim/uvsim.py#L249-L251
The available memory is estimated in two ways. If the code is run by SLURM, it will look for the corresponding environmental variable that gives the available memory per node. If it's not run by SLURM, it will use `psutil` methods to estimate available virtual memory:
https://github.com/RadioAstronomySoftwareGroup/pyuvsim/blob/af54dec9c65ed3ac31e7d1505a21c030745dffbe/pyuvsim/utils.py#L258-L263
- Fixes E-field analytic beam definition, setting off-diagonal terms in the Jones matrices to 0.

Still to be done:
- ~Benchmarking to compare with the current master branch.~
- ~Re-run reference simulations to verify results.~
 x ~Fix the coarse horizon cut. The preferable way to re-implement this requires an option to do selections on `SkyModel` objects, but this has been difficult (as discussed in the channel)~ [This will be omitted for now, kept as an issue, and restored in a future version]
- ~Run on other clusters.~
- ~Write more tests for `get_avail_memory`.~
- ~Enable syntax for more cluster management systems in `get_avail_memory`.~ [Adding a comment in the docs to request other system support via issues.]

closes #196


aelanman:
Reinstating the coarse horizon cut has proven to be far more complicated that it's worth, so I'm going to leave it as an issue for the future.

Remaining things to be done:

- Test this on a different cluster architecture. Shared memory depends a little on how jobs are split across nodes.
- Write syntax for different cluster management systems (grid engine?, others?) into the get_avail_memory() function.

mkolopanis:
I'm more than happy to test on the cluster at ASU if you provide me with the configuration file of a sim you'd like to use as a comparison

aelanman:
Thanks @mkolopanis ! As a first step, could you check that the tests pass? Then maybe try running ref sim 1.4?

mkolopanis:
interestingly I found that you don't necessarily need the `.git` extension on remote url in a git config and my version was failing as a result. but yeah things look good. I'll check out ref sim 1.4

aelanman:
Matt has run the tests, a reference sim, and a benchmarking simulation on enterprise and produced the same visibilities to with 10^-9. The benchmarking sim runtime was comparable to the same job specification on Oscar (4 nodes, 50 PUs).

bhazelton:
As discussed on the telecon, this PR should bump the version number to 1.1 (need to update the VERSION file and the changelog)

aelanman:
@bhazelton I bumped up the version number and fixed some other test errors from the rebase.

CHANGELOG.md
```diff
@@ -1,18 +1,31 @@
 # Changelog
 
-## [Unreleased]
```
bhazelton: You should leave this line with nothing below it so people know where to add things in future.

CHANGELOG.md
```diff
@@ -1,18 +1,31 @@
...
+## [1.1.0] - 2019-6-14
 
-### Added
+## Added
```
bhazelton: looks like we're using three hashes for these items below (so they're smaller than the version numbers)

aelanman: I've fixed this and the the [Unreleased] marker.

pyuvsim/simsetup.py
```diff
@@ -100,80 +138,83 @@ def array_to_sourcelist(catalog_table, lst_array=None, time_array=None, latitude
...
+        ra = ra[~nonrising]
+        dec = dec[~nonrising]
+
+    ids = catalog_table['source_id']
+    freqs = catalog_table['frequency']
+
+    if coarse_horizon_cut:
+        circumpolar = tans >= 1      # These will have rise/set lst set to nan
+        with warnings.catch_warnings():
+            warnings.simplefilter('ignore', RuntimeWarning)
```
bhazelton: This will catch all runtime warnings, is that what you want? I presume there's a specific warning you're trying to catch, might be good to make this more specific. warnings.filterwarnings can take a message.

pyuvsim/source.py
```diff
@@ -5,50 +5,77 @@
...
 from astropy.units import Quantity
 from astropy.time import Time
 from astropy.coordinates import Angle, SkyCoord, EarthLocation, AltAz
 
+from . import utils as simutils
+
+
+def _skymodel_basesize():
+    """
+    Estimate the memory footprint of a SkyModel with a single source.
```
bhazelton: I don't understand this.

pyuvsim/source.py
```diff
@@ -89,54 +133,54 @@ def coherency_calc(self, time, telescope_location):
...
-            raise ValueError('time must be an astropy Time object. '
-                             'value was: {t}'.format(t=time))
-
         if not isinstance(telescope_location, EarthLocation):
             raise ValueError('telescope_location must be an astropy EarthLocation object. '
                              'value was: {al}'.format(al=telescope_location))
+        # stokes shape = (4, Ncomponents)
+        # ra/dec/hour_angle shape = (1,Ncomponents)
+        # Ionly_mask = (Ncomponents)
+        Ionly_mask = np.sum(self.stokes[1:, :], axis=0) == 0.0
```
bhazelton: The implementation here really confuses me and it's going to be very hard to rebase the work James has been doing on it. Could we instead split the array into polarized and unpolarized and then only call the polarized specific code if there are any polarized sources?

pyuvsim/source.py
```diff
@@ -145,48 +189,45 @@ def alt_az_calc(self, time, telescope_location):
...
-    def pos_lmn(self, time, telescope_location):
-        """
-        calculate the direction cosines of this source at a time & location
+        cirs_source_coord = skycoord_use.transform_to('cirs')
+        tee_ra = simutils.cirs_to_tee_ra(cirs_source_coord.ra, time)
 
-        Args:
-            time: astropy Time object
-            telescope_location: astropy EarthLocation object
+        self.hour_angle = None
```
bhazelton: Where is hour_angle used? I'd prefer to stick with alt/az

pyuvsim/source.py
```diff
@@ -145,48 +189,45 @@ def alt_az_calc(self, time, telescope_location):
...
     def __eq__(self, other):
-        return (np.isclose(self.ra.deg, other.ra.deg, atol=self.pos_tol)
-                and np.isclose(self.dec.deg, other.dec.deg, atol=self.pos_tol)
-                and np.all(self.stokes == other.stokes)
-                and (self.name == other.name)
-                and (self.freq == other.freq))
+        return (np.allclose(self.ra.deg, other.ra.deg, atol=self.pos_tol)
+                and np.allclose(self.stokes, other.stokes)
+                and np.all(self.name == other.name)
+                and np.all(self.freq == other.freq))
```
bhazelton: I notice that the time isn't included here, so the objects can be equal if the times are different. I find that confusing because the methods would give different results in that case.

pyuvsim/source.py
aelanman: `hour_angle` comes up in `coherency_calc` (lines 152-153)

pyuvsim/source.py
aelanman: I think the `__eq__` method was written before `time` was an attribute. I'll add it here.

pyuvsim/source.py
aelanman: This is doing the rotation matrix calculation for all sources and then overwriting the rotations as identity matrices for sources with only stokes-I. For now, it probably makes more sense to do the opposite order. The only change is that the rotation matrices now have an `Ncomponents` axis as the last axis.

Looking at the `coherency_basis_rotation_redux` branch, it seems like it will take some effort to vectorize all of the various functions involved.

bhazelton: Ok. I think you will almost always have many more unpolarized sources, so doing the hard work for only the polarized sources make sense. I agree there will be some work to vectorize the polarized stuff, but keeping the unpolarized ones separate will help, I think. We might want to consider doing a squash on the commits on this branch to help with the rebasing.

pyuvsim/source.py
bhazelton: ah, ok. That's going away with the polarized branch, but we can leave it for now.

aelanman:
@bhazelton I've rebased the `vector_sources` branch and done some aggressive squashing. In particular, I've squashed the `coherency_calc` changes into one commit. I've also tried to clean things up by bringing together changes related to the same file, and squashing attempts at different parallelization schemes or SkyModel object splitting.

If this isn't the preferred organization, I still have the original commit history saved elsewhere and I can overwrite this. I do expect that this will minimize merge conflicts when rebasing the coherency calculation branch.

PR #200, opened 2019-05-09
minor bug in test_utils

aelanman:
Fixes an error if running tests without h5py. Last comment from JOSS reviewer.


aelanman:
@bhazelton This is the last little detail the JOSS reviewer noticed. A test was failing if running without h5py because it was trying to clear a test file that doesn't exist. Once this is merged, it's accepted.

Issue #203, opened 2019-05-21
Running test simulations


aishrimc:
I have installed pyuvsim using `pip install pyuvsim`. On running `mpirun -n 20 python run_param_pyuvsim.py obsparam_1.1.yaml`, I am getting `python: can't open file 'run_param_pyuvsim.py': [Errno 2] No such file or directory`, and the ` obsparam_1.1.yaml` file (obtained via 
`wget https://github.com/RadioAstronomySoftwareGroup/pyuvsim/blob/master/reference_simulations/obsparam_ref_1.1.yaml`) is very different from the file in GitHub. I am attaching the file for your reference(in .txt format).
Could you kindly suggest the correct method to run a simulation?
[obsparam_ref_1.1.txt](https://github.com/RadioAstronomySoftwareGroup/pyuvsim/files/3201841/obsparam_ref_1.1.txt)

aelanman:
Hi @aishrimc. If you've installed pyuvsim via pip, you don't need to include "python" in the command. Please try the following:
```
mpirun -n 20 run_param_pyuvsim.py obsparam_1.1.yaml
```

The script `run_param_pyuvsim.py` should be on your path after installation. I see that the wrong syntax is listed on the readthedocs, so we'll have to correct that.

The wget command you're using is downloading the html page off of github, not the raw yaml file. If you go the page `https://github.com/RadioAstronomySoftwareGroup/pyuvsim/blob/master/reference_simulations/obsparam_ref_1.1.yaml` and click the `raw` button, it will take you to the raw file page. Try wget with that path:
```
wget https://raw.githubusercontent.com/RadioAstronomySoftwareGroup/pyuvsim/master/reference_simulations/obsparam_ref_1.1.yaml
```

bhazelton:
@aishrimc Did @aelanman's suggestions work for you?

jpober:
@aishrimc We're closing this issue because we haven't heard from you, but happy to reopen if you're still experiencing issues.  Just let us know.

aishrimc:
Dear Sir,
As of now I'm not having further issues. Apologies for the late response.

Thank you for your help.

PR #204, opened 2019-05-21
fix quick start command

aelanman:
Recent issue #203 indicated that the example command in Quick Start was wrong.


bhazelton:
@jpober can you review this?

PR #210, opened 2019-06-04
Nose to pytest

### LINKED ISSUE #219: IERS table download
Same problem that pyuvdata had. Travis installs astropy and tries to download IERS tables, which can fail if the usno server is down. We need to add some code to ignore the "out of time range" warnings.

This should probably be done with Matt's nose_to_pytest branch.

mkolopanis:
Converts testing suite from `nose` to `pytest`. 

includes pyuvdata iers fix and closes #219

## Todo
- ~Enforce test ordering with `pytest-ordering`. Specifically for the profiling and mpi tests.~ this plugin is still very alpha level. All the ordering seems to be preserved from `nose` currently. perhaps we can revisit when/if necessary.
- ~Convert tests in #198 after PR is merged.~

## Motivation and Context

Ensures future support of testing by changing to a still maintained test suite.

## Types of changes

- [ ] Version change

## Checklist:


bhazelton:
@mkolopanis we merged the vector sources PR, so you'll need to rebase.

mkolopanis:
I rebased off of vector sources, but this branch has had a lot of rebases and it gets a little messy with all test stuff. A good discerning eye is probably needed for review.

pyuvsim/tests/test_simsetup.py
```diff
@@ -119,10 +118,12 @@ def test_param_flux_cuts():
                                                 + [astropy.io.votable.exceptions.W50] * 10)
     catalog = pyuvsim.simsetup.array_to_skymodel(catalog)
     for sI in catalog.stokes[0, :]:
-        nt.assert_true(np.all(0.2 < sI < 1.5))
+        assert np.all(0.2 < sI < 1.5)
 
 
-def check_param_reader(config_num):
+# parametrize will loop over all the give values
+@pytest.mark.parametrize("config_num", list(range(6)))
```
bhazelton: this is cool!

pyuvsim/tests/test_simsetup.py
```diff
@@ -452,13 +460,7 @@ def test_param_select_bls():
     uv_obj_bls2 = uv_obj_full.select(bls=[(40, 41), (42, 43), (44, 45)], inplace=False, metadata_only=True)
 
     uv_obj_bls.history, uv_obj_bls2.history = '', ''
-
-    nt.assert_equal(uv_obj_bls, uv_obj_bls2)
-
-    # check kwargs passed
-    param_dict['object_name'] = 'foo'
-    uv_obj_full, new_beam_list, new_beam_dict = pyuvsim.initialize_uvdata_from_params(param_dict)
-    nt.assert_equal(uv_obj_full.object_name, 'foo')
```
bhazelton: Are  we sure we don't want these three lines?

pyuvsim/tests/test_utils.py
```diff
@@ -127,34 +127,60 @@ def test_file_namer_extensions():
...
+    elif save_format == 'uvh5':
+        assert ofname + '.uvh5' == expected_ofname
+        os.remove(ofname + '.uvh5')
+    else:
+        assert ofname == expected_ofname
+        shutil.rmtree(ofname)
+
+
+def test_write_error_with_no_format():
+    """Test wirte_uvdata will error if no format is given."""
```
bhazelton: wirte -> write

pyuvsim/tests/test_utils.py
```diff
@@ -127,34 +127,60 @@ def test_file_namer_extensions():
...
-        import h5py     # noqa
-        filing_dict['output_format'] = 'uvh5'
-        expected_ofname = simutils.write_uvdata(uv, filing_dict, return_filename=True)
-        nt.assert_equal(ofname + '.uvh5', expected_ofname)
-        os.remove(ofname + '.uvh5')
-    except ImportError:
-        pass  # No h5py
+
+
+def test_file_forma_in_filing_dict():
```
bhazelton: forma -> format

pyuvsim/tests/test_simsetup.py
mkolopanis: Hm I'm not sure why/how these got taken out. I'll take another look at them

pyuvsim/tests/test_simsetup.py
aelanman: Cool as this is... I don't think it's necessary anymore to run tests for all these different configuration files. Most of the other cases covered by the tests for `parse_frequency_params` and `parse_time_params`. That's why I changed it to just run the first config.

I guess it doesn't really hurt to run more tests, except to slow down travis. We should check more closely to see which configuration cases are still needed.

mkolopanis: Hm it did not seem obvious that these test cases were not being run anymore, in fact a new test case was added during one of my re-bases (e.g. there was an `if config_num==5` area).

aelanman: Odd.. I don't know when that was introduced. Currently on master, the `test_param_reader` function loops over the list `[0]`, so it's only running the zeroth case. The coverage is the same, implying some testing redundancy.

mkolopanis: interesting. I'll look to see if those are indeed not needed.

aelanman: Thanks. I probably should have cleaned that up myself before merging that change, but these cases were tricky to define and I'm nervous about removing them unless we're sure they're not needed anymore.

mkolopanis: looks like we gain 1 additional line of coverage by including config_num=2. let me track that down

mkolopanis: looks like in `config_num=2` we get also get this line at simsetup.py line 1458
```
      if par == 'sig': 

            return AnalyticBeam(model, sigma=float(val)) 
```

PR #212, opened 2019-06-10
Keywords to UVData

aelanman:
Re-opening Nick's PR including my changes to his branch.

## Description

Adds a function make UVData objects from keywords which can also make configuration yaml/csv files (optionally). Also ensures that setup functions don't modify parameter dictionaries internally.

Most significantly, this also means that the setup functions can accept polarization selections. Most of the tests ensuring the proper polarization array are removed to allow for this. Instead, an error is raised if the user attempts to run a `pyuvsim` simulation with an improper polarization array (ie., the error case is moved from setup to the simulator).

## Motivation and Context

Addresses the needs of the hera_sim team for methods to set up UVData objects by a function without writing/reading from config files.

## Types of changes

- [ ] Version change

## Checklist:

New feature checklist:


mkolopanis:
oh it looks like it didn't actually _save_ my comments. let me write them all out again.

pyuvsim/tests/test_simsetup.py
```diff
@@ -223,6 +223,25 @@ def test_param_reader():
...
 
+def test_tele_parser():
+    """
+    Check a variety of cases not already tested by param reader
+    """
+    # check no tele config passed
+    tdict = dict(array_layout=os.path.join(SIM_DATA_PATH, 'test_layout_6ant.csv'))
+    nt.assert_raises(KeyError, pyuvsim.simsetup.parse_telescope_params, tdict)
+    tdict['telescope_location'] = '(-30.72152777777791, 21.428305555555557, 1073.0000000093132)'
+    nt.assert_raises(KeyError, pyuvsim.simsetup.parse_telescope_params, tdict)
```
mkolopanis: Again maybe nit-picky but moving forward it would be nice to make sure the correct errors are raised instead of just the correct type of error. These can easily be implemented in a `with` statement and checking the text of the error.

bhazelton: +1

pyuvsim/uvsim.py
```diff
@@ -339,6 +339,10 @@ def run_uvdata_uvsim(input_uv, beam_list, beam_dict=None, catalog=None, source_l
     rank = mpi.get_rank()
     if not isinstance(input_uv, UVData):
         raise TypeError("input_uv must be UVData object")
+
+    if not ((input_uv.Npols == 4) and (input_uv.polarization_array.tolist() == [-5, -6, -7, -8])):
+        raise ValueError("input_uv must have XX,YY,XY,YX polarization")
```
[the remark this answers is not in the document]
bhazelton: It's very non-trivial! Not sure what you mean exactly, but you can't e.g. convert to (true) Stokes in the visibility space.

pyuvsim/tests/test_simsetup.py
```diff
@@ -442,9 +457,74 @@ def test_param_select_redundant():
 
     uv_obj_red2 = uv_obj_full.compress_by_redundancy(tol=0.1, inplace=False, metadata_only=True)
 
+    uv_obj_red.history, uv_obj_red2.history = '', ''
+
     nt.assert_equal(uv_obj_red, uv_obj_red2)
 
 
+def test_uvdata_keyword_init():
```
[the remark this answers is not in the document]
bhazelton: We should definitely be testing that the values are correct.

pyuvsim/uvsim.py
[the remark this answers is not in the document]
mkolopanis: I'm more thinking if a telescope likes to write out with L and R polarizations. I know enough to know I don't really know how polarizations works but it's not easy to simulate one say x, y linear and rotate the basis into combinations of circular. I wasn't really thinking about stokes, those sound scary.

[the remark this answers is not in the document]
bhazelton: Nope. You cannot convert from x,y to L,R trivially (maybe at all in the visibility space). Which pol is recorded is dictated by the design of the feeds and encoded in the beam. Right now we only support linear pol beams (that could potentially be expanded in future but would be work).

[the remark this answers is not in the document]
mkolopanis: very interesting. Thanks :D

pyuvsim/tests/test_simsetup.py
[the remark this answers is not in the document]
aelanman: All of the separate parameter configurations are tested in the tests for `initialize_uvdata_from_params`. This function only sets up the parameters to pass along to that, and there is another test that confirms that the setup from keywords and setup from params make the same thing.

Nonetheless, I think it's a good idea to double-check here and split this test up, so I'll do that.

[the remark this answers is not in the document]
mkolopanis: My comment about shapes and affected parameters also applies to the `initialize_uvdata_from_params` tests. I tried to look through those but only found cases where shapes or sizes were tested. There's a loop  back test where a file is read, turned into a config, made a file, then a config again. The two configs are compared, which is good, but I don't see one that says are all the config parameters equal to the uvdata parameters. maybe I'm just looking too deeply into nothing though.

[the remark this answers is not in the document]
aelanman: I think I see what you mean. `test_param_reader` generates and compares the task lists, which should have all the same information as the UVData objects. Indirectly, this is comparing the attributes of `hera_uv` and the parameters in the config files in `data/test_config`.

A more direct comparison is definitely warranted, though. I'm adding that to the kwarg setup tests now.

pyuvsim/tests/test_simsetup.py
aelanman: There are many `assert_raises` statements throughout the tests... would it make sense to make this change a part of the nose to pytest conversion? It seems pointless to implement it now if we're just going to replace it with pytest later.

mkolopanis: The statements are almost identical in `nose` and `pytest` honestly.

something like this example from an older build of pyuvdata:
```
    with nt.assert_raises(ValueError) as cm:
        uv2.reorder_pols(order=[3, 2, 1])
    ex = cm.exception  # raised exception is available through exception property of context
    nt.assert_true(ex.args[0].startswith('If order is an index array, it must'))
```
and the permalink
https://github.com/RadioAstronomySoftwareGroup/pyuvdata/blob/63b61f84f354448ae500c6e8d6e50c3329dedcc4/pyuvdata/tests/test_uvdata.py#L1157-L1160

it's very easy to convert this to nose.

aelanman: Okay! I'll make the changes.

pyuvsim/simsetup.py
```diff
@@ -595,20 +662,21 @@ def parse_frequency_params(freq_params):
     Parse the "freq" section of obsparam.
 
     Args:
-        freq_params: Dictionary of frequency parameters
+        freq_params: Dictionary of frequency parameters.
+            See pyuvsim documentation for examples of allowable key combinations.
```
bhazelton: more explicit directions about where to find the info would be better that "pyuvsim documentation".

pyuvsim/simsetup.py
```diff
@@ -691,17 +757,19 @@ def parse_time_params(time_params):
 
     Args:
         time_params: Dictionary of time parameters
+            See pyuvsim documentation for examples of allowable key combinations.
```
bhazelton: same comment

pyuvsim/simsetup.py
```diff
@@ -879,19 +955,23 @@ def initialize_uvdata_from_params(obs_params):
...
-    uvparam_dict['Npols'] = 4
 
-    # Now make a UVData object with these settings built in.
-    # The syntax below allows for other valid uvdata keywords to be passed
-    #  without explicitly setting them here.
+    # Parse polarizations
+    if 'polarization_array' not in uvparam_dict:
+        uvparam_dict['polarization_array'] = np.array([-5, -6, -7, -8])
+    if 'Npols' not in uvparam_dict:
+        uvparam_dict['Npols'] = 4
```
bhazelton: why not just make this the length of the polarization array?

pyuvsim/simsetup.py
```diff
@@ -879,19 +955,23 @@ def initialize_uvdata_from_params(obs_params):
...
     time_dict = param_dict['time']
     uvparam_dict.update(parse_time_params(time_dict))
-    uvparam_dict['Npols'] = 4
 
-    # Now make a UVData object with these settings built in.
-    # The syntax below allows for other valid uvdata keywords to be passed
-    #  without explicitly setting them here.
+    # Parse polarizations
+    if 'polarization_array' not in uvparam_dict:
+        uvparam_dict['polarization_array'] = np.array([-5, -6, -7, -8])
```
bhazelton: you might want to call this defaulting out in the docstring. Otherwise you have to read the code to figure out what the default is.

pyuvsim/simsetup.py
```diff
@@ -911,55 +991,174 @@ def initialize_uvdata_from_params(obs_params):
...
+    param_dict['config_path'] = path_out
+    param_dict['param_file'] = os.path.basename(yaml_filename)
+
+    if write_files:
+        tele_params['array_layout'] = antenna_layout_filename
+        with open(yaml_filename, 'w') as yfile:
+            yaml.dump(param_dict, yfile, default_flow_style=False)
+
+    # If layout and telescope config yaml files are not written out, this will be used
+    # to set the antenna names/numbers/positions.
```
bhazelton: This docstring confuses me. Won't the following lines always be used, regardless of whether the file is written?

aelanman: Outdated comment.. I'll delete it.

pyuvsim/simsetup.py
aelanman: I'll add the url to the readthedocs page/section.

aelanman:
@bhazelton @mkolopanis 
Thanks for your feedback! I've made a new function in tests/__init__ that extends the `nt.assert_raises` syntax to also check the error message. This was the quickest way to replace all instances of `nt.assert_raises`.

I've included URLs to the relevant sections of RTD for the different parameter parser functions and addressed other minor comments.

Issue #68, opened 2018-07-20
HEALPix to catalog creator


jpober:
Need something that takes a HEALPix map and makes a catalog (list of source objects)

jpober:
Closed by #216

PR #216, opened 2019-06-18
Healpix to catalog

dseitova:
Added Healpix to catalog function

## Description
 Added healpix_to_sky and read_hdf5 functions to source.py

## Motivation and Context
 Provides way to convert healpix map to a sky model
 https://github.com/RadioAstronomySoftwareGroup/pyuvsim/issues/68

## Types of changes

- [ ] Version change

## Checklist:

Bug fix checklist:

New feature checklist:

Breaking change checklist:  
- [ ] I have updated the docstrings associated with my change using the [numpy docstring format](https://numpydoc.readthedocs.io/en/latest/format.html).
- [ ] I have updated the documentation to reflect my changes (if appropriate).
- [ ] My change includes backwards compatibility and deprecation warnings (if possible).
- [ ] I have added tests to cover my changes.

Reference simulation update or replacement:
- [ ] If this is a new reference simulation or if the settings files have changed, I have added all the relevant files to the reference_simulations folder.
- [ ] If this is a new reference simulation I have fully documented it in a memo in docs and updated any other docs as appropriate.

Documentation change checklist:
- [ ] Any updated docstrings use the [numpy docstring format](https://numpydoc.readthedocs.io/en/latest/format.html).
- [ ] If this is a significant change to the readme or other docs, I have checked that they are rendered properly on ReadTheDocs. (you may need help to get this branch to build on RTD, just ask!)

Version change checklist:
- [ ] I have updated the [CHANGELOG](https://github.com/RadioAstronomySoftwareGroup/pyuvsim/blob/master/CHANGELOG.md) to put all the unreleased changes under the new version (leaving the unreleased section empty).
- [ ] I have run the appropriate tests for this level of version change (described in the readme under versioning).


aelanman:
You'll need to add a couple tests to cover the new methods before this can be merged. I'd suggest making a small HDF5 file similar to the one I sent you with a handful of pixels set to nonzero values and then check that those are read in correctly (e.g., Nside=32, one frequency channel). We can work on more sophisticated tests in the future.

pyuvsim/source.py
```diff
@@ -241,3 +243,39 @@ def get_size(self):
         Estimate the SkyModel memory footprint in bytes
         """
         return self.Ncomponents * self._basesize
+
+
+def read_hdf5(hdf5_filename, freq_ind):
```
bhazelton: if this is specific to healpix, I'd suggest using that in the function name. Many different things can be represented in hdf5 files, including visibilities!

pyuvsim/source.py
```diff
@@ -7,6 +7,8 @@
 import numpy as np
 import copy
 import sys
+import h5py
+import healpy as hp
```
bhazelton: I'd suggest importing these only in the functions that need them to prevent adding more dependencies.

pyuvsim/source.py
aelanman: This is to read a specific hdf5 format for sky models in `healvis`. It's included mainly because we don't have a well-defined file format yet for HEALPix shells.

bhazelton: That's fine, I'm just objecting to the function name -- it should make it clear that it's for healvis's hdf5 rather than any generic hdf5.

pyuvsim/source.py
```diff
@@ -241,3 +241,41 @@ def get_size(self):
...
+        freqs: frequencies from a hdf5 file
+    """
+    import healpy as hp
+    Nside = hp.npix2nside(hpmap.size)
+    dec, ra = hp.pix2ang(Nside, indices, lonlat=True)
+    dec = Angle(dec, unit='deg')
+    ra = Angle(ra, unit='deg')
+    freq = Quantity(freqs, 'hertz')
+    stokes = np.zeros((4, len(indices)))
+    stokes[0] = hpmap
```
aelanman: We'll need to discuss this some more. We're effectively treating healpix pixels (in Kelvin) as point sources (in Jy), which means we'll need to convert the units along the way. This will depend on the map resolution.

jpober:
Are there any more changes you expect to make?  Or should I go ahead and review?

dseitova:
I'm still working on the test function to test the units of healpix_to_sky function, as you suggested. If you think it's not necessary, then it's all ready for review, all nose tests and checks have passed

jpober:
No, let’s get that in there.  Just tag me when you think it’s ready for review.

aelanman:
This should be changed so that it merges into `master`, not `vector_sources`.

mkolopanis:
This branch will need to be updated to use pytest style of tests now.

.travis.yml
```diff
@@ -30,7 +30,7 @@ install:
 
   # install dependencies
   - conda install numpy scipy astropy six nose pip pyyaml line_profiler pycodestyle coveralls h5py psutil
-  - conda install -c conda-forge mpi4py
+  - conda install -c conda-forge mpi4py healpy
```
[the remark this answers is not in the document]
jpober: We discussed this at the 7/19/19 pyuvsim telecon.  I think we do want healpy here for tests to get run by Travis, but overall we don't want healpy as a dependency.

pyuvsim/tests/test_source.py
```diff
@@ -101,3 +106,49 @@ def test_pol_coherency_calc():
...
+def test_units_healpix_to_sky():
+    Nside = 32
+    beam_area = hp.pixelfunc.nside2pixarea(Nside) * units.sr
+    hpmap, inds, freqs = pyuvsim.source.read_healpix_hdf5(os.path.join(SIM_DATA_PATH, 'test_file.hdf5'), 0)
+    freqs = freqs * units.Hz
+    stokes = (hpmap * units.K).to(units.Jy, units.brightness_temperature(beam_area, freqs))
+    sky = pyuvsim.source.healpix_to_sky(hpmap, inds, freqs)
+    print(sky.stokes[0])
+    print(stokes.value)
+    nt.assert_true(np.allclose(sky.stokes[0], stokes.value))
```
jpober: I'd like to see one more test that takes a healpix map with only one non-zero pixel and compares the visibility to an analytic calculation.

pyuvsim/tests/test_source.py
```diff
@@ -9,9 +9,14 @@
 from astropy.time import Time
 from astropy.coordinates import Angle, SkyCoord, EarthLocation
 from astropy import units
+import healpy as hp
```
jpober: Only tests requiring healpy should import healpy and those tests should be skipped if healpy is not installed.

jpober:
Overall, looks great.  Just a few requested changes to documentation and other items discussed at the 7/19/19 pyuvsim telecon.

pyuvsim/source.py
```diff
@@ -241,3 +241,41 @@ def get_size(self):
...
+    flux densities defined by stokes parameters.
+
+    Args:
+        hpmap: healpix map that you get from reading hdf5 file
+        indices: indices from a hdf5 file
+        freqs: frequencies from a hdf5 file
+    """
+    import healpy as hp
+    Nside = hp.npix2nside(hpmap.size)
+    dec, ra = hp.pix2ang(Nside, indices, lonlat=True)
```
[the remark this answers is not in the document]
aelanman: This is backwards --- it should be `ra, dec = hp.pix2ang(...` if lonlat  is True. Further, the units will be degrees (this should be clarified).

For now we're assuming J2000/ICRS. This should also be clarified (maybe a comment that it's in degrees and rename the variables `ra_j2000` etc.?)

It's also worth noting, going forward, that `healpy` can convert among coordinate systems, but it relies on `pyephem` which is deprecated.

pyuvsim/source.py
```diff
@@ -241,3 +241,41 @@ def get_size(self):
         Estimate the SkyModel memory footprint in bytes
         """
         return self.Ncomponents * self._basesize
+
+
+def read_healpix_hdf5(hdf5_filename, freq_ind):
+    """
+    Read hdf5 files using h5py and get a healpix map, indices and frequencies
```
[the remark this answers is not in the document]
aelanman: The hdf5 format is defined by the class `healvis.sky_model.SkyModel` (unfortunate coincidence in naming).

If we're going to use this hdf5 file format, it definitely needs more documentation.

pyuvsim/source.py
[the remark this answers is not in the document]
jpober: Perhaps the new sky model repository needs to define a healpix shell format.  One required parameter there would need to be the epoch of the map.  But it's probably not essential to add to the hdf5 format you're using unless we think that will stick around for a long time.

aelanman:
Something to consider -- the `healpy` package is under the GNU General Public License, while all of our software is BSD licensed. I'd never noticed this before, and I'm not sure how big of a concern that is... `healpy` really is the best existing Python HEALPix package in my opinion. As I understand it, any work derived from GPL-licensed code must also be released under GPL. I don't know if our usage counts as "derivation" or if simply using the package means we can keep our BSD.

There is a package called `astropy-healpix` under development which is specifically BSD-licensed and, as its name implies, is designed to work more directly with `astropy`. When that's more developed it may make sense to switch to that in the future.

environment.yml
```diff
@@ -14,6 +14,7 @@ dependencies:
   - psutil
   - nose
   - line_profiler
+  - healpy
```
jpober: Can we remove this as a dependency?

pyuvsim/source.py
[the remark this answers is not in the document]
jpober: I'm going to assume we'll take care of this with work being done in pyradiosky, so I can live with it for now.

aelanman:
@jpober I'm seeing the same test fail in Python3 on a different PR, so it's not related to these changes here.

pyuvsim/source.py
bhazelton: fix this after we move the code to pyradiosky.

PR #220, opened 2019-06-26
More modular simsetup

steven-murray:
## Description

This PR makes `simsetup.py` more modular, which makes it a more useful set of tools for external codes that may want to use it.

## Motivation and Context

The primary motivation here is in the context of `hera_sim`, though I will try to be as general as possible. It will also attempt _not_ to change any of the current API (i.e. all tests should pass un-modified).

## Types of changes

- [ ] Version change

## Checklist:

New feature checklist:


pyuvsim/simsetup.py
```diff
@@ -964,6 +964,11 @@ def initialize_uvdata_from_params(obs_params):
...
     uvparam_dict.update(parse_time_params(time_dict))
 
+    # There does not seem to be any way to get polarization_array into uvparam_dict, so
+    # let's add it explicitly.
+    if "polarization_array" in param_dict:
+        uvparam_dict['polarization_array'] = param_dict['polarization_array']
+
     # Parse polarizations
     if 'polarization_array' not in uvparam_dict:
         uvparam_dict['polarization_array'] = np.array([-5, -6, -7, -8])
```
aelanman: Somehow, this default is not being set.

pyuvsim/simsetup.py
```diff
@@ -1042,69 +1047,77 @@ def initialize_uvdata_from_params(obs_params):
...
+        Time of the first integration bin [Julian Date]
+    time_array : ndarray
+        time array [Julian Date]. If this is specified it supersedes values of Ntimes, start_time and integration_time
+    bls : list
+        List of antenna-pair tuples for baseline selection
+    redundant_threshold: float
+        Redundant baseline selection tolerance for selection [meters]
+    antenna_nums : list
+        List of antenna numbers to keep in array
+    polarizations : list
```
aelanman: This keyword was changed and should be updated to reflect that. The test that runs this function should also be updated to use `polarization_array` instead of `polarizations`.

pyuvsim/simsetup.py
```diff
@@ -1042,69 +1047,77 @@ def initialize_uvdata_from_params(obs_params):
...
-            If True, write out the parameter information to yaml files.
-        path_out : str (optional)
-            Path in which to place generated configuration files, if write_files is True.
-            Defaults to current directory.
-        kwargs : dictionary
-            Any additional valid UVData attribute to assign to object.
-
-    Returns:
-        UVData object with zeroed data_array
+    Parameters
```
aelanman: I'm realizing now that this docstring is missing descriptions for the following:

- write_files : bool (default False) : Write out configuration files to generate the same UVData object as the given keywords.
- path_out : str (optional)   Path to write out config files, if used. Defaults to current working directory.
- complete: bool (default False)  Makes full UVData object with all required parameters. [Should probably default to True, for the likely use cases of this function].
- no_autos : bool (default False) Exclude autocorrelations from baselines.

.gitignore
```diff
@@ -112,3 +112,5 @@ docs/index.rst
 
 # exclude project specific files
 pyuvsim/GIT_INFO
+
+\.idea/
```
aelanman: What is this?

CHANGELOG.md
```diff
@@ -12,6 +12,7 @@
 ### Changed
 - unit tests check error messages raised.
 - Polarization selection is now allowed in setup. Will break if incorrect polarization is used in pyuvsim.
+- init_uvdata_out function is more modular.
```
aelanman: I think new entries are supposed to go on top.

pyuvsim/simsetup.py
```diff
@@ -1144,23 +1162,36 @@ def initialize_uvdata_from_keywords(yaml_filename=None, antenna_layout_filename=
 
     extra_kwds = {k: v for k, v in six.iteritems(kwargs) if k in valid_param_names}
 
+    # Strangely enough, the initialize_uvdata_from_params doesn't set the polarizations.
+    # This is probably because they were previously hardcoded into the complete_uvdata
+    # function. However, it doesn't really make sense to put them there, so we put
+    # them here.
```
aelanman: It was a legacy of the original design. `pyuvsim` was originally written to read in a uvfits file and simulate based on that, and ignored the polarization array. The `init_uvdata_out` function was made to set up the output file, and was written long before the parameter option was put in. In a much earlier version such simulations were deprecated and now it's exclusively parameter based, so it makes sense to re-organize.

pyuvsim/uvsim.py
```diff
@@ -270,37 +270,13 @@ def init_uvdata_out(uv_in, source_list_name,
...
-    else:
-        # Note: currently only support a constant spacing of times
-        uv_obj.integration_time = (np.ones_like(uv_obj.time_array, dtype=np.float64)
-                                   * np.diff(np.unique(uv_obj.time_array))[0] * (24. * 60**2))  # Seconds
+    uv_obj = simsetup.complete_uvdata(uv_in, inplace=False)
+
     # add pyuvdata version info
     history += uv_obj.pyuvdata_version_str
 
     # Clear existing data, if any
```
aelanman: This comment can go.

pyuvsim/uvsim.py
```diff
@@ -270,37 +270,13 @@ def init_uvdata_out(uv_in, source_list_name,
...
-    else:
-        uv_obj.channel_width = np.diff(uv_obj.freq_array[0])[0]
-    uv_obj.set_uvws_from_antenna_positions()
-    if uv_obj.Ntimes == 1:
-        uv_obj.integration_time = np.ones_like(uv_obj.time_array, dtype=np.float64)  # Second
-    else:
-        # Note: currently only support a constant spacing of times
-        uv_obj.integration_time = (np.ones_like(uv_obj.time_array, dtype=np.float64)
-                                   * np.diff(np.unique(uv_obj.time_array))[0] * (24. * 60**2))  # Seconds
+    uv_obj = simsetup.complete_uvdata(uv_in, inplace=False)
```
aelanman: This function now just adds to the history string and runs `complete_uvdata`.... if it's okay with you, I'd like to shuffle around some of the calls in `uvsim.py` and rename this function.

Mostly, I just don't like that we now have four functions with names that suggest initializing an empty UVData object (and the docstring here suggests as much too), when it should really only be two. I'd like to rename this to "update_history" or something and briefly change how it's called.

aelanman:
@steven-murray I've pushed two commits to your branch removing the `init_uvdata_out` function in favor of `complete_uvdata`, and updated some of the tests. Feel free to revert them if you don't like the changes.

The `vector_sources` branch was merged recently so you'll need to rebase this before this can be merged.

pyuvsim/simsetup.py
steven-murray: I've fixed this particular problem by checking if the "polarization_array" is set to None (not just if it doesn't exist), since the `_from_keywords` function explicitly passes it (sometimes as `None`). 

I do however find it strange that the code was previously checking for `"polarization_array"` in `uvparam_dict` but there was absolutely no way to get it _into_ `uvparam_dict`: passing it in `obs_params` does not result in it being in `uvparam_dict`. I don't know if there was a plan to do that which never got implemented, or something else. Thus I don't know if the way I've done it is in alignment with the ideas of the group.

aelanman: Ah okay, that explains it.

The polarization_array has only recently become a settable parameter, so I'm not surprised. It was technically settable through the 'kwargs', but it wasn't documented and an explicit option is better.

pyuvsim/simsetup.py
steven-murray: ah yes, fair point. Fixed in 977683d20c201c442eac645ad41d9d7ac865d2a0

pyuvsim/simsetup.py
steven-murray: Oh, should mention its fixed in b88770f040a5c4771dd1faa6ed00f8293a3e8a7c

pyuvsim/simsetup.py
steven-murray: I'm not quite sure what you mean -- all these do have descriptions (pretty close to how you are describing them). Are you suggesting updating the descriptions, or were the descriptions totally hidden for you?

I agree that _usually_ people will want to "complete" the object when using this function, however, the reason I left the default as False was for two reasons: 1) to keep the same behaviour as previously by default (though since the function is so new, it may not matter), and 2) because you can always complete the object afterwards, but its more annoying to _undo_ to the completion. 

I'm still happy to change the default if you think it's better.

.gitignore
steven-murray: ugh. that's my editor's project files -- should I remove it and just manually ignore them? Do we want to keep the .gitignore clean?

pyuvsim/simsetup.py
steven-murray: OK, going to remove most of these comments then :-) (ce6c7e0b973cdc8235f465573d7deb7b469d05d4)

pyuvsim/uvsim.py
steven-murray: This seems to be fixed.

pyuvsim/uvsim.py
steven-murray: That seems perfectly reasonable to me!

.gitignore
bhazelton: You should put that kind of thing in your global .gitignore (so they'll be ignored across all the repos on your machine)

pyuvsim/simsetup.py
```diff
@@ -1012,7 +1017,7 @@ def initialize_uvdata_from_params(obs_params):
     uv_obj.history = ''
 
     # select on object
-    valid_select_keys = ['antenna_nums', 'antenna_names', 'ant_str', 'bls', 'frequencies', 'freq_chans', 'times', 'polarizations', 'blt_inds']
+    valid_select_keys = ['antenna_nums', 'antenna_names', 'ant_str', 'bls', 'frequencies', 'freq_chans', 'times', 'blt_inds']
```
bhazelton: why did polarizations get taken out as a valid selection key?

aelanman: It seemed like a strange option to have, since we're defining a UVData object and not selecting on an existing one. I don't see any reason we can't keep it, I'm just not sure what the use case is.

pyuvsim/simsetup.py
aelanman: Oh sorry... I didn't see those keywords in the docstring. Never mind then!

That's a good point on the difficulty in reversing the operation. We can keep the default to False.

.gitignore
steven-murray: Oh yeah, I'd never thought of doing that before! Fixed in 4be03e7abad384b98118f5e6d47c37a9fcfd43be

pyuvsim/simsetup.py
```diff
@@ -1083,80 +1090,108 @@ def initialize_uvdata_from_params(obs_params):
     return uv_obj, beam_list, beam_dict
 
 
-def initialize_uvdata_from_keywords(yaml_filename=None, antenna_layout_filename=None, array_layout=None, telescope_location=None,
-                                    telescope_name=None, Nfreqs=None, start_freq=None, bandwidth=None, freq_array=None, channel_width=None,
-                                    Ntimes=None, integration_time=None, start_time=None, time_array=None,
-                                    bls=None, antenna_nums=None, antenna_names=None, polarizations=None, no_autos=False,
-                                    redundant_threshold=None, run_check=True, write_files=True, path_out=None, **kwargs):
+def initialize_uvdata_from_keywords(
+        yaml_filename=None, antenna_layout_filepath=None, output_layout_filename=None, array_layout=None,
```
aelanman: I'm realizing that `array_layout` is an ambiguous keyword. Currently, it's only the array layout dictionary, so maybe it should be renamed to `antenna_layout_dict` or something?

pyuvsim/uvsim.py
```diff
@@ -366,16 +297,7 @@ def run_uvdata_uvsim(input_uv, beam_list, beam_dict=None, catalog=None, source_l
...
-            Name of the catalog
-        catalog: np.ndarray in shared memory
-            Immutable source parameters
-        obs_param_file: string
-            Parameter filename if running from config files.
-        telescope_config_file: string
-            Telescope configuration file if running from config files.
-        antenna_location_file: string
-            antenna_location file if running from config files.
+        catalog: array of source.Source objects
```
aelanman: This is incorrect. The catalog is passed in as an ndarray in shared memory (see earlier version of the docstring). The Source class no longer exists.

pyuvsim/simsetup.py
steven-murray: I don't find it terribly ambiguous -- especially with the docstring specifying it quite precisely. Is there extra information engendered by "array_layout" that isn't actually present in the dict? Possibly just calling it `antpos` or `antenna_pos` is more clear, but `array_layout` also makes sense to me.

pyuvsim/uvsim.py
steven-murray: I reverted to the previous docstring.

aelanman:
@bhazelton @jpober I believe this branch is ready for review. I've made a bunch of changes to it so I probably shouldn't be the one to review it.

pyuvsim/simsetup.py
```diff
@@ -1083,80 +1090,109 @@ def initialize_uvdata_from_params(obs_params):
...
-            Path in which to place generated configuration files, if write_files is True.
-            Defaults to current directory.
-        kwargs : dictionary
-            Any additional valid UVData attribute to assign to object.
-
-    Returns:
-        UVData object with zeroed data_array
+    Parameters
+    ----------
+    yaml_filename : str (optional)
```
jpober: For added clarity, perhaps you change this to something line output_yaml_file?  Just to make it painfully obvious that this method doesn't accept input yaml files to initialize a uvdata object, that's another method.

pyuvsim/simsetup.py
```diff
@@ -1083,80 +1090,109 @@ def initialize_uvdata_from_params(obs_params):
...
+        yaml_filename=None, antenna_layout_filepath=None, output_layout_filename=None,
+        array_layout=None,
+        telescope_location=None, telescope_name=None, Nfreqs=None, start_freq=None,
+        bandwidth=None, freq_array=None, channel_width=None, Ntimes=None,
+        integration_time=None, start_time=None, time_array=None, bls=None,
+        antenna_nums=None, antenna_names=None, polarization_array=None, no_autos=False,
+        redundant_threshold=None, write_files=True, path_out=None, complete=False,
+        **kwargs):
     """
     Setup a UVData object from keywords.
```
jpober: This description could probably be a bit more informative, saying something about how it writes a yaml file.

pyuvsim/simsetup.py
```diff
@@ -1324,6 +1388,58 @@ def uvdata_to_config_file(uvdata_in, param_filename=None, telescope_config_name=
...
+        uv_obj.channel_width = np.diff(uv_obj.freq_array[0])[0]
+    uv_obj.set_uvws_from_antenna_positions()
+    if uv_obj.Ntimes == 1:
+        uv_obj.integration_time = np.ones_like(uv_obj.time_array, dtype=np.float64)  # Second
+    else:
+        # Note: currently only support a constant spacing of times
+        uv_obj.integration_time = (np.ones_like(uv_obj.time_array, dtype=np.float64)
+                                   * np.diff(np.unique(uv_obj.time_array))[0] * (24. * 60**2))  # Seconds
+
+    # Clear existing data, if any
```
jpober: These seems like an important feature to document: if you pass this function a UVData object with a data array, the data will be zeroed.

pyuvsim/simsetup.py
aelanman: Good idea. I've changed it.

pyuvsim/simsetup.py
aelanman: I've added an explanation that it will optionally write configs that can be used by `initialize_uvdata_from_params`.

pyuvsim/simsetup.py
aelanman: I've added a comment to the docstring, but I also figured it's probably better to make this function private. It doesn't have any external uses. The only user access to it now is through `initialize_uvdata_from_keywords`, which doesn't start with any data in it anyway.

Maybe I should add a warning if it actually overwrites data? I can imagine a weird situation where that could happen...

PR #222, opened 2019-07-11
Fix installation order

steven-murray:
## Description

This implements the same kinds of changes as those merged into `pyuvdata` recently, as well as more solidified support for dev `conda` environments.

## Motivation and Context

Fixes the following issues:
* Installation being circular, so that fresh environments fail to install `pyuvsim` directly.
* Addition of `environment.yml` to make it easy for `conda` users to install a development environment.

I have *not* included all required packages for development (eg. sphinx) in `environment.yml`, as I don't know them. They should probably be added before merging.

## Types of changes

- [ ] Version change

## Checklist:

Bug fix checklist:


README.md
```diff
@@ -24,19 +24,46 @@ pyuvsim's two primary goals are interferometer simulation accuracy at the level
...
-pip install pyuvdata
-pip install .
-```
-* `pyuvsim` is intended to run on clusters running the linux operating system.
+## Installation
+A user-installation is achieved simply with `pip install pyuvsim`, or to get the 
+bleeding-edge: `pip install https://github.com/RadioAstronomySoftwareGroup/pyuvsim`. 
+This will install all dependencies. 
+
+If you wish to manage dependencies manually, or are developing `pyuvsum` yourself, read
```
adampbeardsley: `pyuvsum` -> `pyuvsim`

.travis.yml
```diff
@@ -29,12 +29,15 @@ install:
...
-  - python setup.py install
+  - conda create -q -n test-environment python=$TRAVIS_PYTHON_VERSION coveralls
+  - source activate test-environment
+  - conda env update -n test-environment -f environment.yml
+  - pip install .
   - python -c "from pyuvdata import version; version.main()"
   - conda list
+  # removed this, as it's installed via conda. However, it may be that we *want* it to
+  # be from github...
+  #- pip install git+https://github.com/RadioAstronomySoftwareGroup/pyuvdata.git
```
adampbeardsley: I would bet using the conda version is better here. But should we have an integration test on pyuvdata to test against pyuvsim like we do for hera_qm and hera_cal?

bhazelton: I think the reason we haven't used the conda version is because sometimes there's very recent development on pyuvdata that pyuvsim needs. But I'm ok with that providing more push to get out small version updates from pyuvdata more frequently (or on an as-needed basis).

We could add testing against pyuvsim on pyuvdata if needed.

bhazelton: @aelanman what's your feeling about installing pyuvdata from conda vs the repo for pyuvsim testing?

bhazelton:
I think this looks reasonable but we need @aelanman  to weigh in -- he's the most active developer on this repo.

bhazelton:
I should note @aelanman that I think this will fix the issues that our reviewer was having with installing pyuvsim.

CHANGELOG.md
```diff
@@ -10,7 +10,7 @@
 
 ### Fixed
 - Bug in how integration_time array is set in setup
-
+- Circular-installation bug
```
aelanman: I think the convention is to put more recent changes higher on the list.

.travis.yml
aelanman: In the past, there have been times we needed to build pyuvsim against a pending branch of pyuvdata, or use a bleeding-edge version that isn't available on conda yet. For instance, I'm planning to add a test in the near future that uses the `ignore_requirements` option on UVBase, which I don't think has been published yet to anaconda, and I'd also like to update pyuvsim to work with the new interp syntax on UVBeam. I don't know how often pyuvdata gets updated on conda.

I think for PRs, we should be testing against the latest merged version of pyuvdata (ie., the repo), and for _releases_ we should be testing against what's been published to conda, since the release is what users will probably install and run.

README.md
steven-murray: Thanks. Fixed in 234179010bd3ce72ff46e09aab3327e4b2dac78d

.travis.yml
steven-murray: Is there a way to specify which version of pyuvdata to test against per-test on travis? I agree with @aelanman I think -- releases should definitely be tested against released versions. It makes sense that PRs may want to sometimes test against dev versions of pyuvdata.

Although, I think a good case could be made that any PR which is to be pulled into master for `pyuvsim` should rely only on released versions of `pyuvdata`. The developer can test it against dev versions on their own system, but before merging to master, it should probably be built against a released version. This also forces the version requirements to be updated sufficiently within the PR itself (eg. if the new `pyuvsim` PR depends on the newest features of `pyuvdata`, then the PR should be blocked until a new release of `pyuvdata` is made, at which point the `pyuvsim` PR should be updated to have `install_requires = {..., "pyuvdata">=new.release.version"}`).

CHANGELOG.md
steven-murray: Ah damnit, got that wrong last time as well. Fixed in be478c0a4e33a6d80fc014aa6a1d256d1dd4e229

mkolopanis:
Looks like this is having IERS issue. Once #210  is merged in that should be resolved

mkolopanis:
This won't necessarily need a re-base I think, but will need to be updated to account for pytest changes now

environment.yml
```diff
@@ -0,0 +1,19 @@
...
+  - PyYAML>=5.1
+  - mpi4py>=3.0.0
+  - astropy>=2.0
+  - pip
+  - psutil
+  - nose
+  - line_profiler
+  - pycodestyle
+  - h5py
+  - pyuvdata
```
mkolopanis: please add pytest-cov to the environment

.travis.yml
steven-murray: What did we get to on this issue?

environment.yml
steven-murray: Done in ad351291a34d9d9d172605361b1e6df7c8ba2647

.travis.yml
steven-murray: As discussed on `pyuvsim` telecon, we're going to go ahead and ensure that `pyuvsim` works with _released_ `pyuvdata` versions, and so this will block PRs from merging until the `pyuvdata` release has the necessary feature/behaviour. 

The key to this working is that `pyuvdata` must be released on a reasonably high cadence, (or we end up with the problem we're currently having with `aipy`). @bhazelton reminded us that the best way to do this is to create version-bump PRs on `pyuvdata`.

.travis.yml
```diff
@@ -29,12 +29,15 @@ install:
...
```
bhazelton: Let's remove this commented out section. There's always the history, we can get back to it if needed.

steven-murray: Done in ec6d28e450cd4c960a4b71072ca4edf36ab09f28

steven-murray:
Do we also want to fix `.travis.yml` line 41/42:
```
  # unclear why this call needs to be different than in pyuvdata...
```
in this PR?

bhazelton:
I think the comment is because we found that it did need to be different to work but couldn't figure out why.

plaplant:
@bhazelton @steven-murray I think it's because pyuvdata has a `setup.cfg` file. `pycodestyle` looks at that for configuration details, and since there aren't any, it defaults to adding `.` as the implicit argument

steven-murray:
@bhazelton, @mkolopanis  uh-oh, another IERS issue.

steven-murray:
@plaplant thanks. I'll add in the `setup.cfg`. Actually, can that be put in `pyproject.toml`?

mkolopanis:
> @bhazelton @steven-murray I think it's because pyuvdata has a `setup.cfg` file. `pycodestyle` looks at that for configuration details, and since there aren't any, it defaults to adding `.` as the implicit argument

funny thing is this is pyuvdata's `setup.cfg`
```
[metadata]
description-file = README.md

[aliases]
test=pytest

[tool:pytest]
addopts = --ignore=scripts
```

so it's not obvious why pyuvdata has a different pycodestyle call

steven-murray:
@plaplant sorry, just saw that it has to be `setup.cfg`. Also, do we want to use `flake8` instead of `pycodestyle`? It has all the same features, _plus_ automatic code complexity checking

plaplant:
@mkolopanis ah, didn't realize that pyuvsim already had a `setup.cfg`. I just ran `pycodestyle --ignore=E501,W503` locally for pyuvsim and it worked fine. Does it not work on travis?

@steven-murray I would be a fan of moving over to `flake8`, but I only ever make unsolicited comments on the repo and don't do actual work, so I'll leave that decision up to the primary developers :)

mkolopanis:
hm does just the existence of the `setup.cfg` mean you don't need the `.`?

plaplant:
Yes, that's been my experience

steven-murray:
lol, that's weird. But OK. I still vote for putting the `--ignore`  in the `setup.cfg` anyway. I can easily change it to `flake8`. Any nay-sayers? @bhazelton ?

bhazelton:
I'm potentially interested in flake-8, but I haven't been able to look at it in detail. I'm ok with switching, but we need to test it.

steven-murray:
I'll make a commit with it changed to that, see if we like it.

steven-murray:
@bhazelton do you like it?

setup.cfg
```diff
@@ -3,3 +3,11 @@ test=pytest
...
 [tool:pytest]
 addopts = --ignore=scripts
+
+[flake8]
+ignore = E501, W503
+max-line-length = 79
+# it's recommended to have max-complexity ~ 18
+max-complexity = 30
+# should probably also have "F" here -- bad imports and variable definitions
+select = B,C,E,W,T4,B9
```
bhazelton: what is complexity a measure of? What do these things in select mean? Should F be added?

steven-murray: explicitly, this: https://en.wikipedia.org/wiki/Cyclomatic_complexity

Basically, you want to keep complexity down to make the code have less places to go wrong. I think the default is about 18. `flake8` will error if the complexity is higher than `max-complexity`. The highest in `pyuvsim` is 30 (so I set it to that to get it to pass). I figure reducing it should be another PR. 

The "Select" identifiers give various kinds of formatting errors/bad practices that should be caught. Error codes are mostly from `pycodestyle`: https://pycodestyle.readthedocs.io/en/latest/intro.html#error-codes, but "F" codes are new in `flake8`: http://flake8.pycqa.org/en/latest/user/error-codes.html, and the "C" code is for complexity. 

In particular, "F" catches a lot of stuff like "module imported but not used". Lots of things fail if we also select for "F", which I think should be fixed, but probably in a separate PR.

aelanman: That's very interesting, and looks to be a useful metric. Do you think the current complexity may be coming from the number of setup options we have? Once the simulation starts, there's pretty much just one path it follows.

bhazelton: Ok, sounds reasonable. Reading through the wikipedia page suggests complexity can be measured at different levels. Is 30 a measure for the entire package or for individual files or something else?

steven-murray: that 30 was for some individual _functions_ . Setting `max-complexity=18`, I get the following output:
```
./pyuvsim/simsetup.py:315:1: C901 'create_mock_catalog' is too complex (21)
./pyuvsim/simsetup.py:557:1: C901 'parse_telescope_params' is too complex (30)
./pyuvsim/simsetup.py:709:1: C901 'parse_frequency_params' is too complex (20)
./pyuvsim/simsetup.py:961:1: C901 'initialize_uvdata_from_params' is too complex (21)
./pyuvsim/simsetup.py:1093:1: C901 'initialize_uvdata_from_keywords' is too complex (19)
./pyuvsim/utils.py:182:1: C901 'write_uvdata' is too complex (19)
```

steven-murray: I propose we create an issue for this and move past it in this PR.

bhazelton: Yes, I agree. I'm happier that it's a measure of functions. I agree those should be more simple, I was worried if it was for too large a unit of code (which might need to be complex to work properly).

aelanman: It makes sense that those functions are complex. They handle several many different setup options. I'm a little surprised that `parse_frequency_params` is too complex but `parse_time_params` isn't.

PR #229, opened 2019-07-22
Formatting changes to satisfy flake8

### LINKED ISSUE #228: flake8 select=F
The select=F option is currently off for flake8, because it makes things fail. This should be fixed.

steven-murray:
## Description

This PR closes #228 , making the code pass numerous formatting checks (mostly unused imports and unused variables). 

One thing I am unsure of is the `__init__.py`, which I've had to use `noqa` for every line that does `from <> import *`. Apparently the right way to get around this is to use `__all__` in every module, but I hate doing that, because you end up forgetting to add your new function to it. I'm happy to leave them specifically commented to skip formatting checks, but happy to hear other opinions.

Another idea which could be discussed here is whether we like `black` formatting. Adding `black` as a pre-commit hook can dramatically simplify a lot of these formatting concerns (you still need to run `flake8` as there are some things that `black` can't automatically fix). I know @plaplant is in favor of this, but will leave it up to people who actually write code in this repo more often to decide.

## Types of changes

- [ ] Version change

## Checklist:


docs/make_index.py
```diff
@@ -25,22 +25,7 @@ def write_index_rst(readme_file=None, write_file=None):
...
-    regex_end = re.compile(end_branch_str)
-    loc_branch_end = re.search(regex_end, readme_md).start()
-    branch_str = readme_md[loc_travis_end:loc_branch_end]
-
-    cover_str = 'https://coveralls.io/repos/github/RadioAstronomySoftwareGroup/pyuvsim/badge.svg'
-    regex_cover = re.compile(cover_str)
-    loc_cover_start = re.search(regex_cover, readme_md).start()
-    loc_cover_end = re.search(regex_cover, readme_md).end()
-
+    pypandoc.convert_file(readme_file, 'md')
```
bhazelton: In my experience this doesn't work. pypandoc doesn't properly handle the badges. You'll need to look at readthedocs to verify this works (you can also test it locally by building the sphinx docs).

I now do this a little differently, and I think more cleanly, in pyuvdata, but it still requires cutting out the badges and replacing them with the ones that work in rst.

steven-murray: So I didn't add this in -- it was there already (just removed an unused variable). Willing to fix it in this PR though, if you like.

I'm actually not sure why we don't just write the README in rst -- github displays this perfectly well, and it means we don't have to do any conversion.

pyuvsim/tests/profile_srcs.py
```diff
@@ -5,18 +5,14 @@
 from __future__ import absolute_import, division, print_function
```
aelanman: I don't remember what this script is used for, but I think it's out of date.

pyuvsim/tests/test_simsetup.py
```diff
@@ -4,20 +4,20 @@
...
-import yaml
 import shutil
-import copy
-from six.moves import range
-import pytest
+
 import astropy
-from astropy.time import Time
-from astropy.coordinates import Angle, SkyCoord, EarthLocation, AltAz
+import numpy as np
```
aelanman: What's the new logic behind the order of import statements?

pyuvsim/tests/test_simsetup.py
```diff
@@ -235,7 +247,8 @@ def test_tele_parser():
     assert bdict == {}
 
     tdict.pop('array_layout')
-    simtest.assert_raises_message(KeyError, 'array_layout must be provided.', pyuvsim.simsetup.parse_telescope_params, tdict)
+    simtest.assert_raises_message(KeyError, 'array_layout must be provided.', pyuvsim.simsetup.parse_telescope_params,
+                                  tdict)
```
aelanman: A more general comment --- Are we now setting a maximum line length?

scripts/run_pyuvsim.py
```diff
@@ -7,15 +7,13 @@
```
aelanman: This script should also be deprecated. The only supported wrapper scripts are `run_param_pyuvsim.py` and `run_profile_pyuvsim.py`. This script will break the MPI functions.

pyuvsim/simsetup.py
```diff
@@ -653,7 +653,8 @@ def parse_telescope_params(tele_params, config_path=''):
     return_dict['antenna_names'] = np.array(antnames.tolist())
     return_dict['antenna_numbers'] = np.array(antnums)
     antpos_enu = np.vstack((E, N, U)).T
-    return_dict['antenna_positions'] = uvutils.ECEF_from_ENU(antpos_enu, *telescope_location) - tele_params['telescope_location']
+    return_dict['antenna_positions'] = uvutils.ECEF_from_ENU(antpos_enu, *telescope_location) - tele_params[
+        'telescope_location']
```
mkolopanis: I fully understand breaking lines for ease of reading but breaking the line on the argument to `tele_params` I actually find more confusing. Can we break before/after the operator? I know PEP has slightly conflicting ideas about this. https://lintlyci.github.io/Flake8Rules/rules/W503.html

current best practice might be break before the operator? 

This same thing happens a few times actually.

pyuvsim/simsetup.py
```diff
@@ -746,7 +747,8 @@ def parse_frequency_params(freq_params):
                 raise ValueError("Either channel_width or Nfreqs "
                                  " must be included in parameters:" + kws_used)
             if sf and ef:
-                freq_params['bandwidth'] = freq_params['end_freq'] - freq_params['start_freq'] + freq_params['channel_width']
+                freq_params['bandwidth'] = freq_params['end_freq'] - freq_params['start_freq'] + freq_params[
+                    'channel_width']
```
mkolopanis: same as line-breaking comment

pyuvsim/simsetup.py
```diff
@@ -771,10 +773,12 @@ def parse_frequency_params(freq_params):
...
         if not sf:
             if ef and bw:
-                freq_params['start_freq'] = freq_params['end_freq'] - freq_params['bandwidth'] + freq_params['channel_width']
+                freq_params['start_freq'] = freq_params['end_freq'] - freq_params['bandwidth'] + freq_params[
+                    'channel_width']
         if not ef:
             if sf and bw:
-                freq_params['end_freq'] = freq_params['start_freq'] + freq_params['bandwidth'] - freq_params['channel_width']
+                freq_params['end_freq'] = freq_params['start_freq'] + freq_params['bandwidth'] - freq_params[
+                    'channel_width']
```
mkolopanis: same as line-breaking comment

pyuvsim/simsetup.py
```diff
@@ -771,10 +773,12 @@ def parse_frequency_params(freq_params):
 
         if not sf:
             if ef and bw:
-                freq_params['start_freq'] = freq_params['end_freq'] - freq_params['bandwidth'] + freq_params['channel_width']
+                freq_params['start_freq'] = freq_params['end_freq'] - freq_params['bandwidth'] + freq_params[
+                    'channel_width']
```
mkolopanis: same as line-breaking comment

pyuvsim/simsetup.py
```diff
@@ -851,7 +855,8 @@ def parse_time_params(time_params):
                 raise ValueError("Either integration_time or Ntimes must be "
                                  "included in parameters: " + kws_used)
             if st and et:
-                time_params['duration'] = time_params['end_time'] - time_params['start_time'] + time_params['integration_time'] * dayspersec
+                time_params['duration'] = time_params['end_time'] - time_params['start_time'] + time_params[
+                    'integration_time'] * dayspersec
```
mkolopanis: same as line-breaking comment

pyuvsim/tests/test_analyticbeam.py
```diff
@@ -202,16 +189,17 @@ def test_gaussbeam_values():
 
     engine = pyuvsim.UVEngine(task)
     engine.apply_beam()
-    altitudes = task.sources.alt_az[0]   # In radians.
-    coherencies = np.real(engine.apparent_coherency[0, 0] + engine.apparent_coherency[1, 1]).astype(float)  # All four components should be identical
+    altitudes = task.sources.alt_az[0]  # In radians.
+    coherencies = np.real(engine.apparent_coherency[0, 0] + engine.apparent_coherency[1, 1]).astype(
+        float)  # All four components should be identical
```
mkolopanis: I think this could be an extension of the line-breaking comment too, but it is a test so not necessarily mission critical.

pyuvsim/tests/test_simsetup.py
```diff
@@ -26,7 +26,8 @@
 herabeam_default = os.path.join(SIM_DATA_PATH, 'HERA_NicCST.uvbeam')
 
 # Five different test configs
-param_filenames = [os.path.join(SIM_DATA_PATH, 'test_config', 'param_10time_10chan_{}.yaml'.format(x)) for x in range(6)]
+param_filenames = [os.path.join(SIM_DATA_PATH, 'test_config', 'param_10time_10chan_{}.yaml'.format(x)) for x in
+                   range(6)]
```
mkolopanis: This may also be a personal preference thing too but it always felt like `for` was a more natural place to break statements like this over the next line. perhaps I am in the minority though.

scripts/run_profile_pyuvsim.py
```diff
@@ -66,7 +66,8 @@
     input_uv.ant_1_array, input_uv.ant_2_array = input_uv.baseline_to_antnums(input_uv.baseline_array)
     ants_new = np.unique(input_uv.ant_1_array.tolist() + input_uv.ant_2_array.tolist())
     input_uv.antenna_numbers = ants_new
-    input_uv.antenna_names = ants_new.astype(str)  # Antnames/numbers are going to be messed up by the baseline selection. Unimportant.
+    input_uv.antenna_names = ants_new.astype(
+        str)  # Antnames/numbers are going to be messed up by the baseline selection. Unimportant.
```
mkolopanis: maybe it would be better here to just move the comment to a separate line above/below the execution so the code doesn't need to be wrapped?

docs/make_index.py
bhazelton: I'm sorry, I put the comment on the wrong line. What I'm objecting to is removing all the previous lines. Or are you saying they weren't being used? If that's the case it's either a mistake and the badges aren't rendering properly on readthedocs or pypandoc is now handling the badges properly.

pyuvsim/tests/test_simsetup.py
bhazelton: The logic is actually in pep8, I've been trying to get it right but clearly haven't found all the cases. 

imports should be ordered by:
first standard lib imports
then related packages
then this package

I think there is a bit weirdness here in that imports from the same library are getting separated in a way that makes it hard for me to read them (e.g. pyuvdata.utils is not next to the UVBeam & UVData imports from pyuvdata).

pyuvsim/tests/test_simsetup.py
bhazelton: My preference is to not have a hard cutoff but to strongly prefer line lengths of 80 or less unless longer lines help a lot with readability. I get that that is harder to stick to because there's no enforcement, but I think that unreasonably forcing line wrapping can substantially hurt readability in some cases. The way I do it is that I have a visual marker in my editor that shows me when I've gone beyond 80 characters and then I can choose to make the line shorter or not as a judgement call. But I'm open to another approach.

pyuvsim/simsetup.py
bhazelton: In pyuvdata, we've chose to break before the operator. We enforce this with our choice of ignore arguments to pydocstyle: `--ignore=E501,W503`

Can we do a similar thing for flake8? These conventions have been thoroughly discussed by the pyuvdata developers and agreed on there.

pyuvsim/tests/test_simsetup.py
aelanman: That's pretty much what I've seen so far. I'm just a bit confused here because it looks like the `numpy` import, arguably the _most_ standard library, is getting pushed down in the list. It looks like this happened automatically.

bhazelton: Let me clarify: what is meant by "standard library" are libraries that are packaged with python.

I agree about numpy, and my approach, which differs from flake8 clearly, is to actually have 4 sections, where I insert things like numpy, yaml, six and maybe astropy into a section that comes just after the standardlib section and before the related packages, where I would put pyuvdata and maybe astropy.

bhazelton: @steven-murray will flake8 error if we order things as I describe above?

docs/make_index.py
steven-murray: Yeah they weren't being used (I was a bit confused as well). They are rendering on readthedocs (on master), so I assume pypandoc is handling them fine now?

pyuvsim/tests/profile_srcs.py
steven-murray: Shall I remove it?

pyuvsim/tests/test_simsetup.py
steven-murray: I don't think flake8 will complain -- my editor makes those PEP8 changes automatically. Flake8 will complain if an import is not at the top of the module (i.e. there's logic above it), but I don't think it really cares about ordering too much.

bhazelton: Ok, then my vote is 4 sections as I described and to have imports from the same library next to each other. What do others think?

docs/make_index.py
bhazelton: Interesting. Ok, as long as they're rending well I'm fine with taking them out.

pyuvsim/tests/test_simsetup.py
steven-murray: I agree with @bhazelton . I think line length should be kept to if it makes sense to do so. I find that _almost always_ it's possible to do that. The one referenced here is an example where I think my editor auto-reformatted it. If I'd done it manually, I would have put all the arguments on a new line, with an indent. This is in conformance with `black`.

steven-murray: OTOH, `flake8` has a strict line-limit (we can probably turn it off). I actually think ~88 columns is nicer. I guess if we really thnk a line is worth being really long, we can whack a `# noqa` on the end of it.

bhazelton: Looks to me like we can ignore flake8 errors as we did for pycodestyle.

pyuvsim/simsetup.py
steven-murray: Yup, so all of these (this and below) were auto-formatted by my editor, and I agree with you guys. I'll fix these up.

scripts/run_pyuvsim.py
steven-murray: Removed in b7e89d70fab47c00273dbfa95fa5f5c82f3af8b0

pyuvsim/simsetup.py
steven-murray: Fixed in b7e89d70fab47c00273dbfa95fa5f5c82f3af8b0

pyuvsim/simsetup.py
steven-murray: Fixed in b7e89d70fab47c00273dbfa95fa5f5c82f3af8b0

pyuvsim/simsetup.py
steven-murray: Fixed in b7e89d70fab47c00273dbfa95fa5f5c82f3af8b0

pyuvsim/simsetup.py
steven-murray: Fixed in b7e89d70fab47c00273dbfa95fa5f5c82f3af8b0

pyuvsim/simsetup.py
steven-murray: Fixed in b7e89d70fab47c00273dbfa95fa5f5c82f3af8b0

pyuvsim/tests/test_analyticbeam.py
steven-murray: Fixed in b7e89d70fab47c00273dbfa95fa5f5c82f3af8b0

pyuvsim/tests/test_simsetup.py
steven-murray: Fixed in b7e89d70fab47c00273dbfa95fa5f5c82f3af8b0

scripts/run_profile_pyuvsim.py
steven-murray: this script was removed as per @aelanman 's advice

pyuvsim/tests/profile_srcs.py
steven-murray: Removed in b7e89d70fab47c00273dbfa95fa5f5c82f3af8b0

docs/make_index.py
steven-murray: I have also confirmed by running sphinx locally, and the badges show fine.

pyuvsim/tests/test_simsetup.py
steven-murray: do you want me to re-order the imports again? To be honest, I don't really see much of a difference between imports like `numpy` and `astropy` and `yaml`: to me it's fine if they're all lumped together. But if other people feel strongly, I can try moving a subset of non-standard libraries into their own little group?

pyuvsim/tests/test_simsetup.py
steven-murray: In b7e89d70fab47c00273dbfa95fa5f5c82f3af8b0 I changed the `max-line-length` to 100, and un-ignored the line limit check. Is this what we want? I can just ignore the limit completely, and remove the redundant `max-line-length` if you prefer.

I do really think that having a line limit is a good idea when multiple people are contributing. We just have to agree on how the lines should be split.

steven-murray:
In 386285ce599d8ec2f200f4c2aba58559699fa00c I made a _lot_ of changes to whitespace, making it close to what running `black` would produce, with a line length of 100. Some things that are different to black: I didn't enforce double-quotes for strings, and I didn't enforce that long argument lists or dicts or plain lists have a single entry per line. I did however make a reasonable effort to make such lists start on a new line with a single indent, with the closing parenthesis on a new line to close it. 

If you don't like the new style (I think it makes it much easier to keep within the line limit in a nice way), I can roll back that commit. If you do like it, we should probably consider using `black`: it's much less work (literally just run `black` at the top level and it's done), and can actually be exactly enforced (we can run `black --check` on travis). 

I'm sorry that so many files were modified in so many places!

Thoughts @bhazelton + @aelanman ?

docs/parameter_files.rst
```diff
@@ -100,18 +100,18 @@ Time
...
+        * (``integration_time``, ``Ntimes``)
+        * (``duration_hours``) or (``duration_days``)
+
+    Time step:
+        * (``duration_hours`` or ``duration_days``, ``Ntimes``)
+        * (``integration_time``)
+
+    Reference time:
+        * (``start_time``)
+        * (``end_time``)
```
bhazelton: These changes feel like they might mess up the formatting on RTD. Have they been checked there?

paper.bib
```diff
@@ -1,158 +1,154 @@
...
+    doi = {10.1051/aas:1996146},
+    abstract = {Astronomy and Astrophysics, Supplement Series (A\&AS) published data papers, either observational or theoretical, as well as extensive data material forming the basis of papers with astrophysical results},
+    language = {en},
+    number = {1},
+    urldate = {2017-10-09},
+    journal = {Astronomy and Astrophysics Supplement Series},
+    author = {Hamaker, J. P. and Bregman, J. D. and Sault, R. J.},
+    month = may,
+    year = {1996},
+    pages = {137--147}
```
bhazelton: I'm disinclined to change this file. It's for the accepted JOSS paper and I don't see the utility in changing it (it could also possibly mess things up?)

pyuvsim/__init__.py
```diff
@@ -5,13 +5,13 @@
...
-from .utils import *
+from .profiling import *  # noqa
+from .uvsim import *  # noqa
+from .simsetup import *  # noqa
+from .analyticbeam import *  # noqa
+from .antenna import *  # noqa
+from .baseline import *  # noqa
+from .source import *  # noqa
+from .telescope import *  # noqa
+from .utils import *  # noqa
```
bhazelton: I'm not opposed to moving to defining the `__all__` parameter long term, but this seems like a good short term fix.

docs/parameter_files.rst
bhazelton: I just built it on RTD and it's definitely not quite right. Might just need a couple of empty lines added in places.

pyuvsim/analyticbeam.py
```diff
@@ -46,18 +46,23 @@ class AnalyticBeam(object):
...
+        given by `diameter`)
+        * Gaussian : A peak-normalized gaussian function.
+            * If given a `diameter`, then this makes a chromatic beam with FWHMs matching an
+            equivalent Airy disk beam at each frequency.
+            * If given a `sigma`, this makes an achromatic beam with standard deviation set to
+            `sigma`
+            * If given a `sigma`, `ref_freq`, and `spectral_index`, then this will make a
+            chromatic beam
+              with standard deviation defined by a power law: `stddev(f) = sigma * (
+              f/ref_freq)**(spectral_index)`
```
bhazelton: This docstring isn't rendering properly on RTD: https://pyuvsim.readthedocs.io/en/code-format/classes.html#primary-beams

pyuvsim/analyticbeam.py
```diff
@@ -46,18 +46,23 @@ class AnalyticBeam(object):
...
 
     Args:
-        type: (string)
-            Beam type to use
+        type: str, {'uniform', 'airy', 'gaussian'}
```
bhazelton: This doesn't look like something I've seen in the numpy docstring style. Can you point me to a reference?

pyuvsim/antenna.py
```diff
@@ -47,7 +47,9 @@ def get_beam_jones(self, array, source_alt_az, frequency, reuse_spline=True,
         # 2x2 array of Efield vectors in alt/az
 
         # convert to UVBeam az/za convention
-        source_za, source_az = simutils.altaz_to_zenithangle_azimuth(source_alt_az[0], source_alt_az[1])
+        source_za, source_az = simutils.altaz_to_zenithangle_azimuth(
+            source_alt_az[0], source_alt_az[1]
+        )
```
bhazelton: I've seen this approach before and I get that it reduces the number of line changes in diffs. But I don't personally love it. I'm also not sure it's what PEP8 asks for, but I guess it's passing that check.

pyuvsim/antenna.py
```diff
@@ -64,7 +66,7 @@ def get_beam_jones(self, array, source_alt_az, frequency, reuse_spline=True,
                                                  reuse_spline=reuse_spline)
         Ncomponents = source_za.shape[-1]
 
-        # interp_data has shape: (Naxes_vec, Nspws, Nfeeds, 1 (freq),  Ncomponents (source positions))
+        # interp_data has shape: (Naxes_vec, Nspws, Nfeeds, 1 (freq),  Ncomponents (source pos))
```
bhazelton: This feels to me like a case in which a shorter line hurts readability.

pyuvsim/profiling.py
```diff
@@ -8,40 +8,45 @@
...
 
-def set_profiler(func_list=default_profile_funcs, rank=0, outfile_name='time_profile.out', dump_raw=False):
+def set_profiler(func_list=default_profile_funcs, rank=0, outfile_name='time_profile.out',
+                 dump_raw=False):
     """
     Applies a line profiler to the listed functions, wherever they appear in pyuvsim.
 
-    Places a LineProfiler object in the builtins list, and registers its dumping/printing functions to run
+    Places a LineProfiler object in the builtins list, and registers its dumping/printing
+    functions to run
```
bhazelton: add the next line to this one.

pyuvsim/simsetup.py
```diff
@@ -4,29 +4,29 @@
...
+import yaml
+from astropy.coordinates import Angle, SkyCoord, EarthLocation
 from astropy.io.votable import parse
-from astropy.coordinates import Angle, SkyCoord, EarthLocation, AltAz
-
+from astropy.time import Time
+from numpy.lib import recfunctions
 from pyuvdata import UVBeam, UVData
-import pyuvdata.utils as uvutils
+from six.moves import map, range
```
bhazelton: I'd like to see imports from the same library next to each other. I find that *much* more readable.

pyuvsim/simsetup.py
```diff
@@ -58,14 +65,18 @@ def _parse_layout_csv(layout_csv):
...
     with open(filepath, 'w') as lfile:
         lfile.write(header + '\n')
         for i, (e, n, u) in enumerate(antpos_enu):
             beam_id = 0
             name = antenna_names[i]
             num = antenna_numbers[i]
-            line = ("{:" + str(col_width) + "} {:8d} {:8d} {:10.4f} {:10.4f} {:10.4f}\n").format(name, num, beam_id, e, n, u)
+            line = ("{:" + str(col_width) + "} {:8d} {:8d} {:10.4f} {:10.4f} {:10.4f}\n").format(
+                name, num, beam_id, e,
+                n, u)
```
bhazelton: why are these 2 on a separate line?

pyuvsim/simsetup.py
```diff
@@ -199,26 +208,28 @@ def read_votable_catalog(gleam_votable, input_uv=None, source_select_kwds={}, re
...
             Valid options:
             |  lst_array: For coarse RA horizon cuts, lsts used in the simulation [radians]
-            |  latitude_deg: Latitude of telescope in degrees. Used for declination coarse horizon cut.
-            |  horizon_buffer: Angle (float, in radians) of buffer for coarse horizon cut. Default is about 10 minutes of sky rotation.
-            |                  (See caveats in simsetup.array_to_skymodel docstring)
+            |  latitude_deg: Latitude of telescope in degrees. Used for declination coarse
+            horizon cut.
+            |  horizon_buffer: Angle (float, in radians) of buffer for coarse horizon cut.
+            |      Default is about 10 minutes of sky rotation. (See caveats in
+            simsetup.array_to_skymodel docstring)
```
bhazelton: This docstring isn't rendering properly on RTD: https://pyuvsim.readthedocs.io/en/code-format/classes.html#pyuvsim.simsetup.read_votable_catalog

pyuvsim/simsetup.py
```diff
@@ -258,18 +269,20 @@ def read_text_catalog(catalog_csv, input_uv=None, source_select_kwds={}, return_
...
         input_uv: The UVData object for the simulation (needed for horizon cuts)
         source_select_kwds: Dictionary of keywords for source selection.
             Valid options:
             |  lst_array: For coarse RA horizon cuts, lsts used in the simulation [radians]
-            |  latitude_deg: Latitude of telescope in degrees. Used for declination coarse horizon cut.
-            |  horizon_buffer: Angle (float, in radians) of buffer for coarse horizon cut. Default is about 10 minutes of sky rotation.
+            |  latitude_deg: Latitude of telescope in degrees. Used for declination coarse
+            horizon cut.
+            |  horizon_buffer: Angle (float, in radians) of buffer for coarse horizon cut.
+            Default is about 10 minutes of sky rotation.
```
bhazelton: This docstring doesn't render properly on RTD: https://pyuvsim.readthedocs.io/en/code-format/classes.html#pyuvsim.simsetup.read_text_catalog

pyuvsim/simsetup.py
```diff
@@ -324,22 +338,24 @@ def create_mock_catalog(time, arrangement='zenith', array_location=None, Nsrcs=N
...
-                |  `long-line`:  Horizon to horizon line of point sources
-                |  `hera_text`:  Spell out HERA around the zenith
-                |  `random`:  Randomly distributed point sources near zenith
+            |  `triangle`:  Three point sources forming a triangle around the zenith
+            |  `cross`: An asymmetric cross
+            |  `zenith`: Some number of sources placed at the zenith.
+            |  `off-zenith`:  A single source off zenith
+            |  `long-line`:  Horizon to horizon line of point sources
+            |  `hera_text`:  Spell out HERA around the zenith
+            |  `random`:  Randomly distributed point sources near zenith
```
bhazelton: This doesn't render properly on RTD: https://pyuvsim.readthedocs.io/en/code-format/classes.html#pyuvsim.simsetup.create_mock_catalog

pyuvsim/simsetup.py
```diff
@@ -600,22 +624,26 @@ def parse_telescope_params(tele_params, config_path=''):
...
         tele_params['telescope_location'] = uvutils.XYZ_from_LatLonAlt(*telescope_location)
         telescope_name = telconfig['telescope_name']
 
     else:
         # if not provided, get bare-minumum keys from tele_params
         if 'telescope_location' not in tele_params:
-            raise KeyError("If telescope_config_name not provided in `telescope` obsparam section, you must provide telescope_location")
+            raise KeyError(
+                "If telescope_config_name not provided in `telescope` obsparam section, you must "
+                "provide telescope_location")
```
bhazelton: This seems like a weird solution. We already have a two line message, it just needs to be broken at a different place.

pyuvsim/simsetup.py
```diff
@@ -600,22 +624,26 @@ def parse_telescope_params(tele_params, config_path=''):
...
         if 'telescope_name' not in tele_params:
-            raise KeyError("If telescope_config_name not provided in `telescope` obsparam section, you must provide telescope_name")
+            raise KeyError(
+                "If telescope_config_name not provided in `telescope` obsparam section, you must "
+                "provide telescope_name")
```
bhazelton: same as above

docs/parameter_files.rst
steven-murray: Strangely, what I see on RTD in the `code-format` branch is the same as what I see on the `latest` branch. However, adding in a line break makes the normal statements non-bold, so I've done that.

paper.bib
steven-murray: Sure, I'll revert it.

pyuvsim/simsetup.py
```diff
@@ -719,10 +753,10 @@ def parse_frequency_params(freq_params):
         dict of array properties:
             |  channel_width: (float) Frequency channel spacing in Hz
             |  Nfreqs: (int) Number of frequencies
-            |  freq_array: (dtype float, ndarray, shape=(Nspws, Nfreqs)) Frequency channel centers in Hz
+            |  freq_array: (dtype float, ndarray, shape=(Nspws, Nfreqs)) Frequency channel
+            centers in Hz
```
bhazelton: This is rendered very confusingly on RTD. I think the solution is to add the following to docs/conf.py:
```# set this to handle returns section more uniformly
napoleon_use_rtype = False
```

pyuvsim/source.py
```diff
@@ -157,15 +156,19 @@ def coherency_calc(self, telescope_location):
 
             polarized_sources = np.where(~Ionly_mask)[0]
             sinX = np.sin(self.hour_angle)
-            cosX = np.tan(telescope_location.lat) * np.cos(self.dec) - np.sin(self.dec) * np.cos(self.hour_angle)
+            cosX = np.tan(telescope_location.lat) * np.cos(self.dec) - np.sin(self.dec) * np.cos(
+                self.hour_angle)
```
bhazelton: I'd prefer a break just before the minus sign. Easier to read the math that way.

pyuvsim/source.py
```diff
@@ -157,15 +156,19 @@ def coherency_calc(self, telescope_location):
...
 
-            rotation_matrix = np.array([[cosX, sinX], [-sinX, cosX]]).astype(float)       # (2, 2, Ncomponents)
+            rotation_matrix = np.array([[cosX, sinX], [-sinX, cosX]]).astype(
+                float)  # (2, 2, Ncomponents)
```
bhazelton: Let's put the comment on the line before and then the code won't need to be split.

pyuvsim/analyticbeam.py
steven-murray: https://numpydoc.readthedocs.io/en/latest/format.html#sections

pyuvsim/tests/test_analyticbeam.py
```diff
@@ -109,22 +104,23 @@ def test_uv_beam_widths():
...
+    )
 
     ebeam = interpolated_beam[0, 0, 1, :, :]
     ebeam = ebeam.reshape(Nfreqs, Npix, Npix)
     beam_kern = np.fft.fft2(ebeam, axes=(1, 2))
     beam_kern = np.fft.fftshift(beam_kern, axes=(1, 2))
     for i, bk in enumerate(beam_kern):
-        thresh = np.max(np.abs(bk)) * 0.005  # Cutoff at half a % of the maximum value in Fourier space.
+        thresh = np.max(
+            np.abs(bk)) * 0.005  # Cutoff at half a % of the maximum value in Fourier space.
```
bhazelton: Let's put the comment on the line above so the code doesn't have to be split

pyuvsim/tests/test_analyticbeam.py
```diff
@@ -202,16 +198,19 @@ def test_gaussbeam_values():
 
     engine = pyuvsim.UVEngine(task)
     engine.apply_beam()
-    altitudes = task.sources.alt_az[0]   # In radians.
-    coherencies = np.real(engine.apparent_coherency[0, 0] + engine.apparent_coherency[1, 1]).astype(float)  # All four components should be identical
+    altitudes = task.sources.alt_az[0]  # In radians.
+    coherencies = np.real(
+        engine.apparent_coherency[0, 0] + engine.apparent_coherency[1, 1]
+    ).astype(float)  # All four components should be identical
```
bhazelton: I think this split is caused by the comment. Let's move the comment to before the line to avoid it.

pyuvsim/analyticbeam.py
steven-murray: Oops. I've fixed it, but I also think that that this section should be put inside the parameter description, do you agree?

pyuvsim/analyticbeam.py
bhazelton: Thanks!

pyuvsim/antenna.py
steven-murray: I personally much prefer it. It has the benefit you mentioned, but also means that long lists of arguments are all indented at the same level for all functions called at that level, and also minimizes number of lines used (while staying under a line limit). It's how `black` does it (except it also does a thing which I find a bit annoying: it puts _every_ argument on a new line, if the arguments don't all fit on one, which takes too many lines sometimes).

pyuvsim/antenna.py
steven-murray: sure, I'll fix that.

pyuvsim/profiling.py
steven-murray: haha woops. fixed.

pyuvsim/simsetup.py
steven-murray: absolutely agree. check it in the new commit when I push it.

pyuvsim/antenna.py
bhazelton: I guess the thing that bothers me most is the parenthesis on its own line.

pyuvsim/simsetup.py
steven-murray: damn, missed that. fixed in new commit

pyuvsim/simsetup.py
steven-murray: Thanks, converted that to a list

pyuvsim/simsetup.py
steven-murray: I've tried to change all of these to quoted lists.

pyuvsim/simsetup.py
steven-murray: Sorry how is it weird?

pyuvsim/tests/test_simsetup.py
bhazelton: I'm ok with this as long as imports from the same library are together.

docs/parameter_files.rst
bhazelton: The "total time" line is still appended to the previous line. "Minimum requirements" should maybe also be on it's own line.

pyuvsim/analyticbeam.py
bhazelton: sure, that'd be fine.

pyuvsim/simsetup.py
bhazelton: I think I read the diff wrong. It's ok.

pyuvsim/tests/test_simsetup.py
```diff
@@ -521,7 +612,8 @@ def check_uvdata_keyword_init(case):
 
     elif case == 2:
         # also check that '1' gets converted to [1]
-        # Note -- Currently, pyuvdata's selection does not intersect antenna_nums with bls, but joins them.
+        # Note -- Currently, pyuvdata's selection does not intersect antenna_nums with bls,
+        # but joins them.
```
bhazelton: This comment is out of date, it has been fixed to do intersections on pyuvdata.

scripts/run_profile_pyuvsim.py
```diff
@@ -63,18 +63,23 @@
...
     # Baseline selection:
     input_uv.baseline_array = np.repeat(input_uv.baseline_array[:args.Nbls], args.Ntimes)
-    input_uv.ant_1_array, input_uv.ant_2_array = input_uv.baseline_to_antnums(input_uv.baseline_array)
+    input_uv.ant_1_array, input_uv.ant_2_array = input_uv.baseline_to_antnums(
+        input_uv.baseline_array)
     ants_new = np.unique(input_uv.ant_1_array.tolist() + input_uv.ant_2_array.tolist())
     input_uv.antenna_numbers = ants_new
-    input_uv.antenna_names = ants_new.astype(str)  # Antnames/numbers are going to be messed up by the baseline selection. Unimportant.
+    input_uv.antenna_names = ants_new.astype(
+        str)  # Antnames/numbers are going to be messed up by the baseline selection. Unimportant.
```
bhazelton: put comment on line above to avoid code line split

pyuvsim/simsetup.py
steven-murray: Good idea, I've set those.

pyuvsim/source.py
steven-murray: agreed. fixed.

pyuvsim/source.py
steven-murray: Done.

pyuvsim/tests/test_analyticbeam.py
steven-murray: I put it on the line before, but I still have to split the line.

pyuvsim/analyticbeam.py
```diff
@@ -42,23 +42,19 @@ class AnalyticBeam(object):
...
-    function beam types.
-
     Supported types:
 
         * Uniform beam : Unit response from all directions.
-        * Airy: An Airy disk pattern (the 2D Fourier transform of a circular aperture of width
-        given by `diameter`)
+        * Airy: An Airy disk pattern (the 2D Fourier transform of a circular aperture of
+          width given by `diameter`)
         * Gaussian : A peak-normalized gaussian function.
```
bhazelton: I think you need to take out the space before the colon in this line. The space is causing RTD to render it differently than the rest.

pyuvsim/antenna.py
steven-murray: I like it -- it gives a simple visual clue as to when a function call has ended. I'm happy to move these to the end of the previous line if you/others feel strongly. I don't think there's any consensus in PEP8 or otherwise on this.

In https://community.lsst.org/t/pep8-style-hanging-indentation-and-closing-brackets/618 most people seem to like my style, though they agree that aligning with opening bracket is good too.
This post: https://treyhunner.com/2017/07/craft-your-python-like-poetry/ also seems to prefer this method.

bhazelton: it's ok, I think I'm in the minority on this.

scripts/run_profile_pyuvsim.py
steven-murray: Done

pyuvsim/tests/test_simsetup.py
steven-murray: I removed it

pyuvsim/simsetup.py
```diff
@@ -591,19 +606,22 @@ def parse_telescope_params(tele_params, config_path=''):
...
-            |  antenna_positions: Array of ECEF antenna positions
-            |  telescope_location: ECEF array center location
-            |  telescope_location_lat_lon_alt: Lat Lon Alt array center location
-            |  telescope_config_file: Path to configuration yaml file
-            |  antenna_location_file: Path to csv layout file
-            |  telescope_name: observatory name
-        beam_list:  Beam models in the configuration.
-        beam_dict:  Antenna numbers to beam indices
+        param_dict : dict
+            * `Nants_data`: Number of antennas
```
bhazelton: I think an empty line before this line would help

pyuvsim/analyticbeam.py
steven-murray: Yup, did that.

bhazelton:
Found one more RTD weirdness:

Screen Shot 2019-08-02 at 12 31 28 PM

steven-murray:
Strangely enough, that formatting appears on RTD but not when I make the docs locally. I've shifted that whole paragraph into the notes on the parameters, which is where I think it belongs. We'll have to wait and see the online results, but local results look good.

PR #232, opened 2019-08-05
Fix an issue where values is not defined

steven-murray:
When `type` is uniform and `beam_type` is power, the `values` array was not defined.

### DIFF
--- a/pyuvsim/analyticbeam.py
+++ b/pyuvsim/analyticbeam.py
@@ -132,6 +132,10 @@ class AnalyticBeam(object):
             interp_data[0, 0, 1, :, :] = 1
             interp_data[1, 0, 1, :, :] = 0
             interp_data[0, 0, 0, :, :] = 0
+
+            # If beam_type == "power", we need to know the shape of "values"
+            values = interp_data[0, 0, 0]
+
             interp_basis_vector = None
         elif self.type == 'gaussian':
             if (self.diameter is None) and (self.sigma is None):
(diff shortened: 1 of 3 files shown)


aelanman:
Looks good, but I think you should modify `test_power_analytic_beam` in `test_analyticbeam.py` to break if this fix isn't in place.

bhazelton:
This looks reasonable to me. @aelanman do you agree?

steven-murray:
Sorry I never got time to fix the test. Let me try do that quickly.

steven-murray:
Aaaaand its failing on our two favourite errors: IERS and 19 != 20.

steven-murray:
I've tried restarting now 3 times, and each time it breaks with a different combination of errors. This last time, it was 66 != 20...

aelanman:
I propose that we tell pytest to skip the counter test until we figure out what's going wrong with it. It's testing a very minor feature that's provided for convenience and it shouldn't hold up important developments.

bhazelton:
Can we turn this counter test into an issue so we don't lose track of it?

mkolopanis:
This issue exists already. Is it what you want? #235 i saw the word `counter` and immediate assumed it was this one. If not sorry to bust in.

aelanman:
That issue is about this problem. I spent some time last Friday trying to fix it to no avail and raised the issue to keep track of it.

steven-murray:
Not sure why coveralls isn't reporting. Can we just merge?

mkolopanis:
Interesting, the coverage is on coveralls if you follow the link from travis, but starting at the coveralls main page for the repo, it is not listed under "recent builds".

steven-murray:
What the hell. Coveralls is so flaky sometimes.

bhazelton:
we can just merge, I'll tweak the settings to do it.

PR #233, opened 2019-08-28
Installation without MPI and other tools

steven-murray:
## Description

This PR gives the ability to install `pyuvsim` without requiring `MPI` or `psutils`, on the observation that these extra packages are only needed for actual _simulation_, not for all the other useful tools that `pyuvsim` provides. 

It makes installing these extra options easy, via the `"extras_require"` keyword in `setup()`. 
Along with these simulation-specific packages, there were a few existing optional dependencies listed in the README which have now been formally included as extras in `setup()`, making it easy
to get everything needed for running `pyuvsim`.

## Motivation and Context

Closes #191 

## Types of changes

- [ ] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)
- [ ] Reference simulation update or replacement
- [ ] Documentation change (documentation changes only)
- [ ] Version change
- [x] Build or continuous integration change

Build or continuous integration change checklist:
- [x] If required or optional dependencies have changed (including version numbers), I have updated the readme to reflect this.
- [x] If this is a new CI setup, I have added the associated badge to the readme and to references/make_index.py (if appropriate).

PR #234, opened 2019-08-30
Allow module import when mpi4py not installed

### LINKED ISSUE #191: importing pyuvsim.simsetup without needing mpi
The setup tools should be accessible to other simulation packages, but accessing them requires from pyuvsim requires that mpi4py be installed, which can be difficult. We should either migrate these functions to a new repository, or allow for pyuvsim to be installed without mpi4py.

mkolopanis:
wrap imports for optional dependencies in try/except blocks, only raise errors when they are used.

## Description

uses try/catch statements to import `mpi4py` and only errors if a function which requires mpi4py is called.

## Motivation and Context

Other repos want to be able to install and use simsetup features without needing mpi4py.
closes #191 

## Types of changes

- [ ] Version change

## Checklist:

Bug fix checklist:

### DIFF
--- a/pyuvsim/profiling.py
+++ b/pyuvsim/profiling.py
@@ -15,9 +15,9 @@ import six
 
 import pyuvsim as _pyuvsim
 try:
-    from .mpi import start_mpi, get_rank
+    from . import mpi
 except ImportError:
-    raise ImportError("You must install mpi4py to use the profiling module")
+    mpi = None
 
 try:
     from line_profiler import LineProfiler
@@ -45,7 +45,10 @@ def set_profiler(func_list=default_profile_funcs, rank=0, outfile_name='time_pro
         outfile_name: Filename for printing profiling results.
         dump_raw: Write out a pickled LineStats object to <outfile_name>.lprof (Default False)
     """
-    start_mpi()
+    if mpi is None:
+        raise ImportError("You need mpi4py to use the uvsim module. Install pyuvsim[sim]")
+
+    mpi.start_mpi()
     global prof
     prof = LineProfiler()
     if prof is None:  # pragma: no cover
@@ -63,7 +66,7 @@ def set_profiler(func_list=default_profile_funcs, rank=0, outfile_name='time_pro
                         prof.add_function(item)
 
     # Write out profiling report to file.
-    if get_rank() == rank:
+    if mpi.get_rank() == rank:
         ofile = open(outfile_name, 'w')
         atexit.register(ofile.close)
         atexit.register(prof.print_stats, stream=ofile)
--- a/pyuvsim/profiling.py
+++ b/pyuvsim/profiling.py
@@ -46,13 +46,17 @@ def set_profiler(func_list=default_profile_funcs, rank=0, outfile_name='time_pro
         dump_raw: Write out a pickled LineStats object to <outfile_name>.lprof (Default False)
     """
     if mpi is None:
-        raise ImportError("You need mpi4py to use the uvsim module. Install pyuvsim[sim]")
+        raise ImportError("You need mpi4py to use the profiling module. "
+                          "Install it by running pip install pyuvsim[sim] "
+                          "or pip install pyuvsim[all] if you also want h5py "
+                          "and line_profiler installed.")
 
     mpi.start_mpi()
     global prof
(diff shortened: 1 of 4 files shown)


mkolopanis:
This seems to be an issue with `counter`, but that is unrelated to anything in this PR. @aelanman any ideas? I can see this locally too but it comes/goes with each build, sometimes counts to 19, sometimes 17, 0, or passes.

aelanman:
Hm... strange that it's only happening in Python 2. I can't get the test to break locally, though. I installed the pytest-counter plugin and ran the test a hundred times at once. It's possible that a newer version of something is letting it break.

The counter works by running a thread on the rank 0 process. It's possible that pytest is killing the thread somehow.

How is your environment set up?

mkolopanis:
I have been able to do it with a fresh environment updated with the environment.yml in the repo. If you're not concerned we don't have to dig too deeply though.

mkolopanis:
also it being a python 2 thing makes me not too worried about it

aelanman:
It's worrying that it's only happening intermittently, and especially that it's happening on the PR build only. My understanding is that that means that the breaking change is in `master` now.

It may not be a big issue (the counter isn't exactly a critical function) but it can be annoying to occasionally have an unrelated test break. We could make the test optional?

mkolopanis:
Yeah I can see how that is not ideal. The imports in those modules do a try/except and raise an importError right now. I believe the idea was that in things like `uvsim` you shouldn't be calling it if you don't have mpi4py installed anyway, otherwise every function would have to raise errors individually? Doesn't sound fun, but we could do that.

mkolopanis:
also there is a pytest plugin that we could use to mark as optional and maybe run if you the user wants https://pypi.org/project/pytest-optional-tests/

mkolopanis:
@bhazelton this approach is more similar to how `psutils` was handled in #233

pyuvsim/profiling.py
```diff
@@ -45,7 +45,10 @@ def set_profiler(func_list=default_profile_funcs, rank=0, outfile_name='time_pro
         outfile_name: Filename for printing profiling results.
         dump_raw: Write out a pickled LineStats object to <outfile_name>.lprof (Default False)
     """
-    start_mpi()
+    if mpi is None:
+        raise ImportError("You need mpi4py to use the uvsim module. Install pyuvsim[sim]")
```
bhazelton: This may not be transparent to users who aren't used to the extras_require syntax. What about:
> You need mpi4py to use the uvsim module. Install it by running `pip install pyuvsim[sim]` or `pip install pyuvsim[all]` if you also want h5py and line_profiler installed.

The other places with similar language should also be updated.

pyuvsim/profiling.py
```diff
@@ -45,11 +45,18 @@ def set_profiler(func_list=default_profile_funcs, rank=0, outfile_name='time_pro
...
+                          "or pip install pyuvsim[all] if you also want h5py "
+                          "and line_profiler installed.")
+
+    mpi.start_mpi()
     global prof
     prof = LineProfiler()
     if prof is None:  # pragma: no cover
-        raise ImportError("line_profiler module required to use profiling tools.")
+        raise ImportError("You need line_profiler to use the profiling module. "
+                          "Install it by running pip install pyuvsim[all]. ")
```
bhazelton: What about combining these two error messages? If you want to run this module you'll need both and you'll need to use the [all] option.

pyuvsim/utils.py
```diff
@@ -261,7 +261,10 @@ def get_avail_memory():
     using psutils methods.
     """
     if not HAVE_PSUTIL:
-        raise ImportError("You need psutils to run this function. Install pyuvsim[sim]")
+        raise ImportError("You need psutils to estimate available memory. "
+                          "Install it by running pip install pyuvsim[sim] "
+                          "or pip install pyuvsim[all] if you also want mpi4py, "
+                          "h5oy, and line_profiler installed.")
```
bhazelton: I think mpi4py will get installed if you use the [sim], so that's extra here. Also `h5oy` should be `h5py`

pyuvsim/profiling.py
mkolopanis: yeah good point.

PR #238, opened 2019-09-05
Setting freq interp

aelanman:
## Description

Fixes UVBeam interp calls according to newest pyuvdata changes.

## Motivation and Context

`UVBeam.interp()` cannot be called with `freq_interp_kind` anymore. This fixes the call in pyuvsim accordingly and ensures that the correct frequency interpolation is set before interp is called.

## Types of changes

- [ ] Version change

## Checklist:

Bug fix checklist:


bhazelton:
we have a good coveralls report, but the status check isn't coming through for some reason. Merging anyway.

PR #105, opened 2018-09-06
Coherency basis rotation redux: DO NOT MERGE

jaguirre:
Brought in Zac's code and did some parallel testing in a notebook.  Looks like obvious bugs are caught.  ~There is an open question about both azimuth and za/alt conventions for the signs.~

Need to add tests of the new calculation in source.py.

Fixes #2

PR #239, opened 2019-09-13
Properly handle polarized sources

### LINKED ISSUE #2: need coherency_local calc
Need a matrix to rotate a Jones matrix from RADec to AzEl.

bhazelton:
## Description

Fix the local coherency calculation for polarized sources.

## Motivation and Context

The current calculation is incorrect. This PR replaces #105 which was too out of date to rebase.
Closes #2 

## Types of changes

## Checklist:

Bug fix checklist:

### DIFF
--- a/pyuvsim/source.py
+++ b/pyuvsim/source.py
@@ -8,6 +8,7 @@ import sys
 
 import six
 import numpy as np
+from scipy.linalg import orthogonal_procrustes as ortho_procr
 from astropy.coordinates import Angle, SkyCoord, EarthLocation, AltAz
 from astropy.time import Time
 from astropy.units import Quantity
@@ -163,7 +164,6 @@ class SkyModel(object):
         ''' This transformation matrix is generally not orthogonal
             to better than 10^-7, so let's fix that. '''
 
-        from scipy.linalg import orthogonal_procrustes as ortho_procr
         R_screwy = axes_altaz.cartesian.xyz
         R_really_orthogonal, _ = ortho_procr(R_screwy, np.eye(3))
 
@@ -277,11 +277,10 @@ class SkyModel(object):
             # If there are any polarized sources, do rotation.
             rotation_matrix = self._calc_coherency_rotation(telescope_location)
 
-            for src_ind in range(self.Ncomponents):
-                coherency_local[:, :, src_ind] = np.einsum(
-                    'ab,bc,cd->ad', rotation_matrix[:, :, src_ind].T,
-                    self.coherency_radec[:, :, src_ind],
-                    rotation_matrix[:, :, src_ind])
+            coherency_local = np.einsum(
+                'xab,bcx,cdx->adx', rotation_matrix.T,
+                self.coherency_radec,
+                rotation_matrix)
 
         # Zero coherency on sources below horizon.
         coherency_local[:, :, self.horizon_mask] *= 0.0
@@ -320,12 +319,6 @@ class SkyModel(object):
 
         time.location = telescope_location
         lst = time.sidereal_time('apparent')
-
-        cirs_source_coord = skycoord_use.transform_to('cirs')
-        tee_ra = simutils.cirs_to_tee_ra(cirs_source_coord.ra, time)
-
-        self.hour_angle = None
-        self.hour_angle = (lst - tee_ra).rad
 
         self.time = time
         alt_az = np.array([source_altaz.alt.rad, source_altaz.az.rad])
--- a/pyuvsim/source.py
+++ b/pyuvsim/source.py
@@ -318,7 +318,6 @@ class SkyModel(object):
                                                        location=telescope_location))
 
         time.location = telescope_location
-        lst = time.sidereal_time('apparent')
 
         self.time = time
         alt_az = np.array([source_altaz.alt.rad, source_altaz.az.rad])
--- a/pyuvsim/source.py
+++ b/pyuvsim/source.py
@@ -6,6 +6,7 @@ from __future__ import absolute_import, division, print_function
 
 import sys
 
+import six
 import numpy as np
 from astropy.coordinates import Angle, SkyCoord, EarthLocation, AltAz
 from astropy.time import Time
--- a/pyuvsim/source.py
+++ b/pyuvsim/source.py
@@ -205,7 +205,7 @@ class SkyModel(object):
         for src_i in range(self.Ncomponents):
             intermediate_vec = np.matmul(R_avg, radec_vec[:, src_i])
 
-            R_perturb = scbt.vecs2rot(intermediate_vec, altaz_vec[:, src_i])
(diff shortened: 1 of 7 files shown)


aelanman:
I've run a bunch of profiling simulations on Oscar with this branch and haven't noticed any performance hit. Evidently these changes don't affect existing performance (for unpolarized sources).

bhazelton:
great, thanks for checking that @aelanman

pyuvsim/source.py
```diff
@@ -35,6 +37,29 @@ class SkyModel(object):
 
     Used to calculate local coherency matrix for source brightness in AltAz
     frame at a specified time.
+
```
aelanman: I'm not sure how, but it looks like the docstring for the __init__ function got moved here.

pyuvsim/source.py
```diff
@@ -112,17 +114,137 @@ def __init__(self, name, ra, dec, freq, stokes,
...
+                             location=telescope_location, frame='icrs',
+                             **rep_dict)
+
+        axes_altaz = axes_icrs.transform_to('altaz')
+        setattr(axes_altaz, rep_keyword, 'cartesian')
+
+        ''' This transformation matrix is generally not orthogonal
+            to better than 10^-7, so let's fix that. '''
+
+        from scipy.linalg import orthogonal_procrustes as ortho_procr
```
aelanman: Should this be moved to the top?

pyuvsim/source.py
```diff
@@ -149,26 +275,13 @@ def coherency_calc(self, telescope_location):
...
-            rotation_matrix = np.array([[cosX, sinX], [-sinX, cosX]]).astype(float)
-            rotation_matrix = rotation_matrix[..., polarized_sources]
-
-            rotation_matrix_T = np.swapaxes(rotation_matrix, 0, 1)
-            coherency_local[:, :, polarized_sources] = np.einsum(
-                'abx,bcx,cdx->adx', rotation_matrix_T,
-                self.coherency_radec[:, :, polarized_sources],
-                rotation_matrix
-            )
+            for src_ind in range(self.Ncomponents):
```
aelanman: If I'm not mistaken, this part is already vectorized on the master branch and the same syntax can be used here:

```
            polarized_sources = np.where(~Ionly_mask)[0]
            rotation_matrix_T = np.swapaxes(rotation_matrix, 0, 1)
            coherency_local[:, :, polarized_sources] = np.einsum(
                'abx,bcx,cdx->adx', rotation_matrix_T,
                self.coherency_radec[:, :, polarized_sources],
                rotation_matrix
            )
```

pyuvsim/source.py
```diff
@@ -202,7 +315,8 @@ def update_positions(self, time, telescope_location):
             return
 
         skycoord_use = SkyCoord(self.ra, self.dec, frame='icrs')
-        source_altaz = skycoord_use.transform_to(AltAz(obstime=time, location=telescope_location))
+        source_altaz = skycoord_use.transform_to(AltAz(obstime=time,
+                                                       location=telescope_location))
```
aelanman: Below here, there's a transformation to cirs and then to tee_ra which is used to calculate the source hour angles for all sources, which was used by coherency rotation. It looks like that functionality is not used anymore, so we can probably get rid of it and avoid doing an unnecessary transform.

pyuvsim/spherical_coordinates_basis_transformation.py
```diff
@@ -0,0 +1,295 @@
...
+# Copyright (c) 2018 Radio Astronomy Software Group
+# Licensed under the 3-clause BSD License
+
+from __future__ import absolute_import, division, print_function
+
+import numpy as np
+
+
+def r_hat(theta, phi):
+    """
```
aelanman: This is equivalent to the `ang2vec` function in `healpy`, but doesn't appear to be available yet in astropy-healpix. In astropy, there's `astropy.coordinates.UnitSphericalRepresentation`.

Not strictly important, but I figure it's worth mentioning.

pyuvsim/source.py
bhazelton: I moved it on purpose. The numpy style says that `__init__` docstrings should go on the object. This is important because methods that start with underscores don't get built automatically using sphinx.

aelanman: Oh I see now. Sorry, I misread the diff and thought it had been moved to the `_skymodel_basesize` function instead.

pyuvsim/source.py
bhazelton: Yes, I think so since SciPy is listed as a required dependency.

pyuvsim/source.py
bhazelton: I don't think that part of the code is run in the master branch. Trying to make that change errors for me.

bhazelton: Ignore the above, I just hadn't figured out ellipses in `einsum`. The real problem is that the `rotation_matrix` is different per source. I plan to vectorize it in a future PR.

aelanman: It should already be a separate rotation matrix per source (the "x" axis in the summation string). The problem may be that you have Ncomponents rotation matrices when the einsum only expects them for the polarized sources.
So you'd replaced `rotation_matrix` with `rotation_matrix[:,:,polarized_sources]` and likewise for `.rotation_matrix_T`.

bhazelton: The Ncomponents is ok because I'm currently calculating the rotation matrix for all the sources, not just the polarized ones. The real problem was that the transposition puts the source axis first.

pyuvsim/source.py
bhazelton: yeah, we don't need hour angle. I'll take that out.

pyuvsim/spherical_coordinates_basis_transformation.py
bhazelton: ok, I haven't looked at `astropy.coordinates.UnitSphericalRepresentation`. Are you suggesting we use that instead of this?

bhazelton: Hmm, that does look like it might do what we want. I'll test with it a little. @jaguirre you might like to know about this too.

aelanman: If this works, I don't see the harm in using it. It might also be faster than calling an external library, depending on how it's used. I ran some short tests and it looks like the astropy method is somewhat slower:
```
Astropy : 3.39 ms ± 0 ns per loop (mean ± std. dev. of 1 run, 1 loop each)
r_hat : 232 µs ± 0 ns per loop (mean ± std. dev. of 1 run, 1 loop each)
```

bhazelton: I looked at it more, it doesn't quite do what we want anyway. It only gives the unit vectors along lat & lon, not along r, and lat is 90 degrees off of theta, which we use throughout this code. So given that it's fast, I think we should stick with what we have.

PR #251, opened 2019-12-16
More robust config path

steven-murray:
## Description

Makes sure the input path is an abspath before splititng it, otherwise you can get an empty string for the dirname, which errors.

## Motivation and Context

Otherwise passing in "myfile.yaml" breaks.

## Types of changes

- [ ] Version change

## Checklist:

Bug fix checklist:


steven-murray:
@aelanman I think this is ready for review -- really just a single line change (plus a test).

PR #252, opened 2019-12-27
Multiple analytic beams

### LINKED ISSUE #187: Beam parameters assigned to beam_ids
Currently, it is not possible to assign, say, different diameters to different beam_ids within a telescope_config yaml. For instance,
```
beam_paths:
  0: 'airy'
  1: 'gaussian'
sigma: 0.3
diameter=14.0
```

This will assign diameter=14m to both the airy and gaussian beams. There's currently no way to specify which analytic beam parameters go to which beams.

One way to solve this is to change the syntax so that each beam path is a list:

```
beam_paths:
  0: 'airy', diameter=15.0
  1: 'gaussian', diameter=14.0
```

aelanman:
## Description

Provides the means to specify diameter, sigma, etc. inline within the "beam_paths" list for telescope configuration files, rather than setting a single global sigma/diameter, etc. For instance:
```
beam_paths:
   0 : airy, diameter=14
   1 : airy, diameter=20
   2 : airy
   3 : gaussian, sigma=0.3
diameter: 30
```
The above will make airy beams for IDs 0, 1, and 2, with diameters 14, 20, and 30 meters respectively. Beam 3 will be a gaussian with sigma parameter 0.3. The "inline" parameters override the global parameter (diameter : 30).

Unsupported parameters are currently ignored.

## Motivation and Context

Currently, analytic beams may only be defined with one global shape parameter. In order to use various analytic beams across the array, we need to be able to specify different properties for each one.

closes #187 

## Types of changes

- [ ] Version change

## Checklist:

New feature checklist:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,9 +2,17 @@
 
 ## [Unreleased]
 
+### Added
+- Support for individual shape parameters for analytic beams.
+
+### Changed
+
+### Fixed
+- Removed warning catch for astropy GLEAM read in tests -- new versions don't raise the warning.
+
 ## [1.1.1] - 2019-11-14
 
-## Added
+### Added
 - A function for checking the memory usage on each Node
 
 ### Changed
--- a/docs/parameter_files.rst
+++ b/docs/parameter_files.rst
@@ -139,7 +139,7 @@ Telescope Configuration
 
     .. literalinclude:: example_configs/bl_lite_mixed.yaml
 
-    This yaml file provides the telescope name, location in latitude/longitude/altitude in degrees/degrees/meters (respectively), and the `beam dictionary`. In this case, beam_id == 0 is the UVBeam file written out to hera.uvbeam, and beam_id == 1 is an Airy beam with diameter 16m. The dictionary only needs to be as long as the number of unique beams used in the array, while the layout file specifies which antennas will use which beam type. This allows for a mixture of beams to be used, as in this example.
+    This yaml file provides the telescope name, location in latitude/longitude/altitude in degrees/degrees/meters (respectively), and the `beam dictionary`. In this case, beam_id == 0 is the UVBeam file hera.uvbeam, beam_id == 1 is an Airy disk with diameter 14 m, beam_id == 2 is a Gaussian beam with sigma 0.03, and beam_id == 3 is another Airy beam. When no shape parameter is written inline in the beam_dictionary (as with 3), pyuvsim will look for a default parameter below. So in this case, the beam_id == 3 ends up with a diameter of 12 m. The dictionary only needs to be as long as the number of unique beams used in the array, while the layout file specifies which antennas will use which beam type. This allows for a mixture of beams to be used, as in this example. Unassigned beams will be ignored (the given layout file does not use beams 2 or 3).
 
     Analytic beams may require additional parameters.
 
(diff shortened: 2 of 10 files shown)


CHANGELOG.md
```diff
@@ -2,9 +2,17 @@
 
 ## [Unreleased]
 
+### Added
+- Support for individual shape parameters for analytic beams.
+
+### Changed
+
+### Fixed
+- Removed warning catch for astropy GLEAM read in tests -- new versions don't raise the warning.
```
bhazelton: Yay! I'm glad they fixed this, it was really an error on their end.

docs/parameter_files.rst
```diff
@@ -139,15 +139,15 @@ Telescope Configuration
...
 
     Analytic beams may require additional parameters.
 
     - uniform = The same response in all directions. No additional parameters.
     - gaussian = Gaussian function shaped beam. Requires either an antenna diameter (in meters) or a standard deviation sigma (in radians). This standard deviation sets the width of the beam in zenith angle. Note that defining gaussian beams via `sigma` will be deprecated in the future.
```
bhazelton: This may be out of scope for this PR, but does the impeding deprecation of sigma mean that we won't be able to specify non-chromatic gaussian beams in future? If so, we might want to rethink that.

pyuvsim/tests/test_mpi_uvsim.py
```diff
@@ -88,12 +88,15 @@ def test_run_paramfile_uvsim():
...
-    )
+    if six.PY2:
+        pyuvsim.uvsim.run_uvsim(param_filename)
+    else:
+        uvtest.checkWarnings(
+            pyuvsim.uvsim.run_uvsim, [param_filename],
+            nwarnings=1,
+            message=['The default for the `center` keyword has changed'],
+            category=[DeprecationWarning]
+        )
```
bhazelton: I'd suggest using pytest to filter out this warning (put this right above the test: `@pytest.mark.filterwarnings("ignore:The default for the `center` keyword has changed")`) unless you really want to check that it is raised. It's going away soon in pyuvdata (the PR is close to merging) and doing it this way also introduces more six dependence which we're going to remove shortly.

docs/parameter_files.rst
aelanman: I've never been entirely clear on why we want to remove support for `sigma`. It seems like a perfectly sensible way of defining a power beam.
There is an option to give a power law evolution to the Gaussian beam width, but only if the width at a given frequency is specified by a `sigma`.

aelanman:
@bhazelton I'm not sure what's happening now... setuptools is failing. Is it just that we can't test in python 2 anymore?

The python3 tests should be passing. I replaced the `center` keyword warning catches with a global filter on them.

aelanman:
@bhazelton Since we're dropping python 2 support in the next PR anyway, I just removed the python2 checks from this PR rather than go hunting for the setuptools error. This reduces coverage a little, but not in unexpected ways.

pyuvsim/tests/conftest.py
```diff
@@ -41,3 +42,9 @@ def setup_and_teardown_package():
     # clean up the test directory after
     if os.path.exists(testdir):
         shutil.rmtree(testdir)
+
+
+@pytest.fixture(autouse=True)
+def ignore_center_deprecation():
+    warnings.filterwarnings('ignore', message='The default for the `center` keyword has changed.',
+                            category=DeprecationWarning)
```
bhazelton: I'm going to make an issue to remove this after the pyuvdata v1.5 branch is merged.

PR #264, opened 2020-02-05
BeamList class

aelanman:
## Description

Adds a new `BeamList` class to telescope.py, which can hold the set of beam objects or the identifying strings for beam objects, as well as some global parameters for using UVBeams.

## Motivation and Context

The beam setup has been inelegant. `simsetup` parses the information from the telescope yaml file and turns it into strings, which another function later interprets in order to initialize the beam objects (either AnalyticBeam or UVBeam). This is done because UVBeam objects can be large and shouldn't be scattered through MPI.

With these changes, a lightweight object replaces the list of strings, and carries methods to convert beam objects to strings or vice versa. It also carries a dictionary of extra parameters for UVBeams, such as `freq_interp_kind` and `interpolation_function`, which users may want to set.

closes #238 

## Types of changes

- [ ] Version change

## Checklist:

New feature checklist:

### DIFF
--- a/pyuvsim/telescope.py
+++ b/pyuvsim/telescope.py
@@ -46,10 +46,10 @@ class BeamList(object):
     """
     A container for the set of beam models and related parameters.
     """
-    float_params = {'sig' : 'sigma', 'diam' : 'diameter',
-                    'reff' : 'ref_freq', 'ind' : 'spectral_index'}
+    float_params = {'sig': 'sigma', 'diam': 'diameter',
+                    'reff': 'ref_freq', 'ind': 'spectral_index'}
 
-    uvb_params = {'freq_interp_kind' : 'cubic',
+    uvb_params = {'freq_interp_kind': 'cubic',
                   'interpolation_function': 'az_za_simple'}
 
     string_mode = True
@@ -94,7 +94,7 @@ class BeamList(object):
         if beam_model.startswith('analytic'):
             bspl = beam_model.split('_')
             model = bspl[1]
-            
+
             to_set = {}
             for extra in bspl[2:]:
                 par, val = extra.split('=')
@@ -137,7 +137,7 @@ class BeamList(object):
 
     def set_str_mode(self):
         if not self._obj_beam_list == []:
-            ## Convert object beams to string definitions
+            # Convert object beams to string definitions
             self._str_beam_list = [self._obj_to_str(bobj) for bobj in self._obj_beam_list]
         self._obj_beam_list = []
         self.string_mode = True
--- a/pyuvsim/telescope.py
+++ b/pyuvsim/telescope.py
@@ -45,9 +45,61 @@ class Telescope(object):
 class BeamList(object):
     """
     A container for the set of beam models and related parameters.
+
+    Each rank in the simulation gets a copy of the set of beam objects
+    used for calculating Jones matrices. Rather than broadcasting the objects
+    themselves, the list is composed of strings which provide either a complete
+    description of an analytic beam or the path to a beamfits file. After the
+    broadcast, the beams are initialized from these strings.
+
+    This class provides methods to transform between the string and object
+    representations, and also carries parameters used globally by all UVBeams,
+    including their frequency / angular interpolation options.
+
+    This behaves just as a normal Python list in terms of indexing, appending, etc.
+
+    Attributes
+    ----------
+    string_mode : bool
+        Is True if the beams are represented as strings, False if they're objects
+
+    uvb_params : dict
+        Set of additional attributes to set on UVBeam objects.
+
+    Parameters
+    ----------
+    beam_list : list (optional)
+        A list of UVBeam/AnalyticBeam objects or strings describing beams. If
+        beam_list consists of objects, then the BeamList will be initialized in
+        object mode with string_mode == False. If beam_list consists of strings,
+        the BeamList will be initialized in string mode.
+
+        Passing in a mixture of strings and objects will error.
+
+    uvb_params : dict (optional)
+        Options to set uvb_params, overriding settings from passed-in UVBeam objects.
+
+    Raises
+    ------
+    ValueError :
+        For invalid beam_list, or if UVBeams in the input beam_list have
+        conflicting parameters (e.g., one has freq_interp_kind == 'cubic' and
+        another has freq_interp_kind == 'linear').
+
+    Notes
+    -----
+    When a UVBeam is appended to the beam list, its parameters override any in
+    uvb_params.
+
+    If an object beam is added while in string mode, it will be converted to a string
+    before being inserted, and vice versa. ***In order to be converted to a string, a UVBeam
+    needs to have the entry 'beam_path' in its extra_keywords, giving the path to the beamfits
+    file it came from.***
+
     """
-    float_params = {'sig': 'sigma', 'diam': 'diameter',
-                    'reff': 'ref_freq', 'ind': 'spectral_index'}
+
+    _float_params = {'sig': 'sigma', 'diam': 'diameter',
+                     'reff': 'ref_freq', 'ind': 'spectral_index'}
 
     string_mode = True
 
@@ -73,7 +125,19 @@ class BeamList(object):
         self.uvb_params.update(uvb_params)    # Optional parameters for the UVBeam objects
 
     def _update_uvb_params(self, ind=slice(None)):
-        # Ensure that uvb_params are consisten with existing UVBeams.
+        """
+        Updates the uvb_params according to the parameters
+        of any UVBeams in the list.
+
+        This will check to ensure that the parameters are consistent
+        among UVBeams.
+
+        Parameters
+        ----------
+        ind : int or slice (optional)
+            Choose a beam object, or set of them, to read for parameters.
+
+        """
 
         if self.string_mode:
             return
@@ -92,12 +156,17 @@ class BeamList(object):
                 raise ValueError('Conflicting settings for {}: {}'.format(key, str(svals)))
 
     def append(self, value):
+        """
+        Append to the beam list, converting objects to strings if in object mode,
+        or vice versa if in string mode.
+        """
         if self.string_mode:
             self._str_beam_list.append(self._obj_to_str(value))
             self._update_uvb_params(ind=-1)
         self._obj_beam_list.append(self._str_to_obj(value))
 
     def __len__(self):
+        # Note that only one of these lists has nonzero length at a given time.
         return len(self._obj_beam_list) + len(self._str_beam_list)
 
     def __getitem__(self, ind):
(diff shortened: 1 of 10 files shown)


bhazelton:
I'm having a hard time understanding what is going on in the code. Can you add detailed bumpy-style docstrings to each class and method?

pyuvsim/telescope.py
```diff
@@ -37,3 +40,243 @@ def __eq__(self, other):
...
+
+    Each rank in the simulation gets a copy of the set of beam objects
+    used for calculating Jones matrices. Rather than broadcasting the objects
+    themselves, the list is composed of strings which provide either a complete
+    description of an analytic beam or the path to a beamfits file. After the
+    broadcast, the beams are initialized from these strings.
+
+    This class provides methods to transform between the string and object
+    representations, and also carries parameters used globally by all UVBeams,
+    including their frequency / angular interpolation options.
```
bhazelton: I think, based on code below, that these parameters are *required* to match between different UVBeams. If that's the case, that needs to be made more explicit. What happens in practice if you pass in UVBeam objects with different settings for these parameters and don't set a global override?

pyuvsim/telescope.py
```diff
@@ -37,3 +40,243 @@ def __eq__(self, other):
...
+    ------
+    ValueError :
+        For invalid beam_list, or if UVBeams in the input beam_list have
+        conflicting parameters (e.g., one has freq_interp_kind == 'cubic' and
+        another has freq_interp_kind == 'linear').
+
+    Notes
+    -----
+    When a UVBeam is appended to the beam list, its parameters override any in
+    uvb_params.
```
bhazelton: Does this mean that my uvb_params on all my UVBeam objects are set by the last one I added? That seems a little perverse.

pyuvsim/telescope.py
```diff
@@ -37,3 +40,243 @@ def __eq__(self, other):
...
+    file it came from.***
+
+    """
+
+    _float_params = {'sig': 'sigma', 'diam': 'diameter',
+                     'reff': 'ref_freq', 'ind': 'spectral_index'}
+
+    string_mode = True
+
+    def __init__(self, beam_list=None, uvb_params={}):
```
bhazelton: I think you don't want to set uvb_params to an empty dict, that can give very strange and unexpected behavior (see e.g. https://docs.python-guide.org/writing/gotchas/). I think you should set it to None and then handle that case explicitly.

pyuvsim/telescope.py
bhazelton: I think if you make the list from objects that don't match you want to get a very clear error or warning message.

pyuvsim/tests/test_simsetup.py
```diff
@@ -225,8 +225,10 @@ def test_param_reader(config_num):
 
     # Check default configuration
     uv_obj, new_beam_list, new_beam_dict = pyuvsim.initialize_uvdata_from_params(param_filename)
-    for i, bm in enumerate(new_beam_list):
-        new_beam_list[i] = pyuvsim.simsetup.beam_string_to_object(bm)
+    new_beam_list.set_obj_mode()
+    print(new_beam_list)
+#    for i, bm in enumerate(new_beam_list):
+#        new_beam_list[i] = pyuvsim.simsetup.beam_string_to_object(bm)
```
bhazelton: These lines can probably be removed.

pyuvsim/tests/test_simsetup.py
```diff
@@ -804,14 +806,14 @@ def test_multi_analytic_beams():
     #   0 : airy, diameter=14
     #   1 : airy, diameter=20
     #   2 : gaussian, sigma=0.5
-
+    print(pyuvsim.BeamList()._str_beam_list)
```
bhazelton: this line can probably be removed.

pyuvsim/telescope.py
aelanman: That is essentially the behavior I was going for, but in hindsight I think there's a better option. The issue is that different parameters among UVBeams is lost when converted to strings, since those parameters are stored collectively in the `uvb_params` dictionary.

A safer option might be to allow for mismatches among UVBeams when in object mode, and raise an error in that case when converting to string mode. I'll have it raise a warning when adding a non-matching UVBeam to a list of objects. I'm not sure what to do when adding an UVBeam in string mode... that UVBeam's parameters will either have to overwrite the existing `uvb_params` or will be lost.

pyuvsim/telescope.py
aelanman: Huh... I never knew that! Thanks, I'll switch it out.

pyuvsim/telescope.py
```diff
@@ -37,3 +40,243 @@ def __eq__(self, other):
...
+            if isinstance(bm, UVBeam):
+                for key, val in self.uvb_params.items():
+                    setattr(bm, key, val)
+
+    def append(self, value):
+        """
+        Append to the beam list, converting objects to strings if in object mode,
+        or vice versa if in string mode.
+
+        If in string mode, this will override the uvb_params if the new object
```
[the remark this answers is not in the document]
aelanman: I forgot to finish writing it... It's fixed now.

pyuvsim/telescope.py
aelanman: Agreed. I've changed the behavior -- In object mode, conflicts in UVBeam parameters are allowed, but it will warn you if you try to add a UVBeam that conflicts. If you try to convert to string mode, it will error if there are conflicts.

Essentially, a BeamList in object mode behaves exactly like a list of beam objects -- it doesn't care what's on them -- but since differences in, e.g., `interpolation_function` will be lost by converting to strings, it won't let you if such differences exist.

pyuvsim/tests/test_simsetup.py
aelanman: Removed. Thanks for catching that.

pyuvsim/tests/test_simsetup.py
aelanman: Done.

aelanman:
@bhazelton I've changed the default behavior to be less destructive (see one of my comments above). It should work now to raise errors/warnings when appropriate, and will not overwrite parameters on the objects given to it.

pyuvsim/telescope.py
```diff
@@ -37,3 +41,278 @@ def __eq__(self, other):
...
+            self._scrape_uvb_params(self._obj_beam_list, strict=False)
+
+    def append(self, value):
+        """
+        Append to the beam list, converting objects to strings if in object mode,
+        or vice versa if in string mode.
+
+        If in string mode, this will overwrite the uvb_params with the added
+        object's parameters.
+        """
```
bhazelton: Is this still true? Will it overwrite the params for the existing list?

pyuvsim/telescope.py
```diff
@@ -37,3 +41,278 @@ def __eq__(self, other):
...
+                    else:
+                        warnings.warn(err.args[0])  # Use exception message as warning message.
+
+        new_pars = {k: v for k, v in new_pars.items() if not v == []}
+
+        return new_pars
+
+    def _set_params_on_uvbeams(self, beam_objs):
+        """
+        Set parameters on UVBeam objects using uvb_params.
```
bhazelton: This docstring will error with flake8 when we turn on docstring testing in Steven's PR -- One line docstrings should fit on one line including quotes.

pyuvsim/telescope.py
```diff
@@ -37,3 +41,278 @@ def __eq__(self, other):
...
+
+    def __getitem__(self, ind):
+        if self.string_mode:
+            return self._str_beam_list[ind]
+        return self._obj_beam_list[ind]
+
+    def __setitem__(self, ind, value):
+        """
+        Insert into the beam list, converting objects to strings if in object mode,
+        or vice versa if in string mode.
```
bhazelton: This docstring will error with flake8 when we turn on docstring testing in Steven's PR -- The first line should be a single line sentence (ending in a period). More details should follow in text after a blank line.

pyuvsim/telescope.py
```diff
@@ -37,3 +41,278 @@ def __eq__(self, other):
...
+            self._str_beam_list[ind] = self._obj_to_str(value)
+        else:
+            value = self._str_to_obj(value)
+            self._obj_beam_list[ind] = value
+            self._scrape_uvb_params(self._obj_beam_list, strict=False)
+
+    def append(self, value):
+        """
+        Append to the beam list, converting objects to strings if in object mode,
+        or vice versa if in string mode.
```
bhazelton: This docstring will error with flake8 when we turn on docstring testing in Steven's PR -- The first line should be a single line sentence (ending in a period). More details should follow in text after a blank line.

pyuvsim/telescope.py
```diff
@@ -37,3 +41,278 @@ def __eq__(self, other):
...
+        except KeyError:
+            raise ValueError("Need to set 'beam_path' key in extra_keywords for UVBeam objects.")
+
+        return path
+
+    def set_str_mode(self):
+        """
+        Set string_mode = True, converting object beams to their
+        string representations and putting UVBeam parameters into the
+        uvb_params dictionary.
```
bhazelton: This docstring will error with flake8 when we turn on docstring testing in Steven's PR -- The first line should be a single line sentence (ending in a period). More details should follow in text after a blank line.

pyuvsim/telescope.py
```diff
@@ -37,3 +41,278 @@ def __eq__(self, other):
...
+            new_uvb_par = self._scrape_uvb_params(self._obj_beam_list, strict=True)
+            self.uvb_params.update(new_uvb_par)
+            self._str_beam_list = [self._obj_to_str(bobj) for bobj in self._obj_beam_list]
+        self._obj_beam_list = []
+        self.string_mode = True
+
+    def set_obj_mode(self):
+        """
+        Set string_mode = False, initializing beam objects from the strings.
+        Additional parameters on UVBeam objects will be set from the uvb_params dictionary.
```
bhazelton: This docstring will error with flake8 when we turn on docstring testing in Steven's PR -- The first line should be a single line sentence (ending in a period). More details should follow in text after a blank line.

pyuvsim/telescope.py
aelanman: This is no longer true, thanks for catching that.

aelanman:
I've updated the docstrings for the flake8 check. (I should probably update my own vim flake8 plugin to check for that too.)

jsdillon:
Hey @bhazelton @adampbeardsley @steven-murray looks like this PR removed the function `beam_string_to_object` that hera_sim was using, causing hera_sim and hera_cal tests to fail: https://travis-ci.org/HERA-Team/hera_sim/jobs/639615288

What should we do here?

aelanman:
@jsdillon Where in hera_sim is it used? I can try to fix it there.

Sorry about that... I didn't realize anyone else was using that function. This PR replaced that behavior. Instead, pyuvsim now uses a BeamList class which can represent beams as either strings or objects.

jsdillon:
https://github.com/HERA-Team/hera_sim/blob/30cbc391ecbdd2a56f9c684902a3a2e41dfee691/hera_sim/visibilities/simulators.py

aelanman:
I've opened a PR that should fix it, but it looks like something else is failing. [https://github.com/HERA-Team/hera_sim/pull/82](https://github.com/HERA-Team/hera_sim/pull/82)

jsdillon:
I think I fixed that. Let's see if that takes care of it

PR #269, opened 2020-03-04
Reference simulation 1

### LINKED ISSUE #260: Benchmarking script
We want a button that will send a standard benchmarking simulation job to a designated supercomputer with established resources, to confirm that a change does not drastically slow down the simulation. This resembles a test, but is only run when deemed necessary by developers/admins.

aelanman:
## Description

Adds some new tools to simplify the process of running the reference simulations and comparing their output to a standard set of files, hosted externally.

This should be merged before #273 

## Motivation and Context

Our PR template says that major changes must be verified by rerunning the reference simulations and comparing to an older run. This older set of data is not easy to find, buried on lustre. I have put the latest verified reference simulation results in a publicly-accessible place, and added a script here that downloads them. These changes also provide a wrapper and job script for running the reference simulations on a SLURM cluster and organizing the results with some metadata for better record-keeping.

Closes #260 

## Types of changes

- [ ] Version change

## Checklist:

New feature checklist:
 (in the README of the reference_simulations directory)
(I compared this latest iteration of the reference simulations to an earlier run. This should, ideally, be the last set of version 1 reference sims).


reference_simulations/README.md
```diff
@@ -6,5 +6,27 @@ reference simulations have grown in complexity. Each set is detailed in its own
...
+        This launches SLURM jobs, described by the script `jobscript.sh`.
+ - jobscript.sh
+        This setups up an output directory, named by the current date, into which output files,
+   slurm output files, profiling data, and records of which job IDs correspond with each obsparam file.
+ - compare_with_last.py
+        Given the paths to the latest output files (uvh5), this will compare the data in those files
+   to the corresponding files in `latest_ref_data`.
+ - download_latest_data.py
+        Run this first to download the reference sim data from the google drive. The new data should be
+   compared with these files. They will end up in a directory `latest_ref_data`.
```
bhazelton: maybe swap these two script descriptions since the second one needs to be done first?

reference_simulations/compare_with_last.py
```diff
@@ -0,0 +1,37 @@
...
+import sys
+import glob
+import warnings
+import os
+
+import h5py
+
+# Loading data using h5py directly instead of reading the full uvh5 files.
+#   (The 1.2 reference sims recalculate all lsts when reloaded, which takes a long time.)
+
```
bhazelton: This recalculation can be turned off if the `lst_array` is already in the file by setting `run_check_acceptability=False`. But if they're not in the file there's no way to turn it off.

reference_simulations/compare_with_last.py
```diff
@@ -0,0 +1,37 @@
...
+new_files = dict(zip(new_file_names, new_file_paths))
+old_files = dict(zip(old_file_names, old_file_paths))
+
+for k in new_files.keys():
+    if k not in old_files.keys():
+        warnings.warn("File {} not in reference data.".format(k))
+        continue
+    with warnings.catch_warnings():
+        warnings.simplefilter('ignore')     # Ignore known_telescopes warning
```
bhazelton: If you're using h5py directly, I don't think there will be a `known_telescopes` warning.

reference_simulations/README.md
aelanman: Done.

reference_simulations/compare_with_last.py
aelanman: Thanks for the tip! With that in place, I can use pyuvdata instead of h5py, so I've switched to that.

reference_simulations/compare_with_last.py
aelanman: With the switch back to using uvdata, the warning catcher is needed again.

aelanman:
@bhazelton I don't know what's happening with that coverage check... but I think this is ready now.

reference_simulations/compare_with_last.py
```diff
@@ -29,9 +25,14 @@
...
         warnings.warn("File {} not in reference data.".format(k))
         continue
+    uv_old = UVData()
+    uv_new = UVData()
     with warnings.catch_warnings():
         warnings.simplefilter('ignore')     # Ignore known_telescopes warning
-        uv_old = h5py.File(old_files[k], 'r')['Data']
-        uv_new = h5py.File(new_files[k], 'r')['Data']
+#        uv_old = h5py.File(old_files[k], 'r')['Data']
+#        uv_new = h5py.File(new_files[k], 'r')['Data']
```
bhazelton: Delete these commented lines (and the one 3 lines below) unless there's a compelling reason to keep them

bhazelton:
The coverage is not a concern -- you deleted some print lines that were covered, so total coverage percent dropped but the delta (change in number of uncovered lines) is zero.

PR #262, opened 2020-02-03
Description of comparisons. Better than nothing.

dannyjacobs:
## Description
A basic Readme for the comparisons directory. It could be better.

## Motivation and Context
Explains why the reference simulations exist and gives some basic information about how to reproduce them.
closes #167 
## Types of changes
- [ ] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)
- [ ] Reference simulation update or replacement
- [x] Documentation change (documentation changes only)
- [ ] Version change
- [ ] Build or continuous integration change

## Checklist:

No checklists are relevant.

PR #270, opened 2020-03-06
Update CI setup ala pyuvdata and pyradiosky

bhazelton:
## Description

Adopt what we've learned on pyuvdata and pyradiosky -- use github actions, including testing on Mac & Linux. Add a minimum dependency run and a pre-commit hook.

The pre-commit hook raised an issue with our yaml formatting for analytic beams, so I changed that convention to make it actually yaml compatible. I also moved to using `pytest.raises` rather than our own custom function.

## Types of changes

- [ ] Version change

## Checklist:


bhazelton:
The github actions all finished before the Travis, even though it's doing quite a bit more total computation:
github actions: (3 python versions * 2 os versions + 1 min dependencies + 1 linting)  * 2 for push/pull = 16 

Travis: (3 python versions * 2 for tests/linting) * 2 for push/pull = 12

bhazelton:
I don't know why coveralls isn't reporting on this PR though...

reference_simulations/Description/reference_simulations_1_summary.tex
```diff
@@ -12,7 +12,7 @@
 
 \title{pyuvsim Reference Simulations \\ version 1}
```
aelanman: This file will be removed with PR #262 , since this description tex file is redundant with the markdown file.

.gitignore
```diff
@@ -112,3 +112,4 @@ docs/index.rst
 
 # exclude project specific files
 pyuvsim/GIT_INFO
+pyuvsim/data/temporary_test_data/*
```
steven-murray: This is fine, but I should note that `pytest` has a fixture for creating proper `tmp` directories, which we should probably migrate to.

.travis.yml
```diff
@@ -28,6 +28,7 @@ install:
   - conda create -q -n test-environment python=$TRAVIS_PYTHON_VERSION coveralls
   - source activate test-environment
   - conda env update -n test-environment -f environment.yml
+  - conda install flake8
```
steven-murray: shouldn't have to install flake8 -- can just install pre-commit and run `pre-commit run -a`.

ci/min_deps.yaml
```diff
@@ -0,0 +1,18 @@
+name: min_deps
+channels:
+  - conda-forge
+  - defaults
+dependencies:
+  - astropy>=4.0
+  - coverage
+  - h5py
+  - numpy>=1.15
```
steven-murray: If we're testing min deps, shouldnt we be enforcing installation of the minimum deps, rather than saying anything above a certain min?

pyuvsim/mpi.py
```diff
@@ -75,7 +75,7 @@ def shared_mem_bcast(arr, root=0):
     itemsize = 0
     dtype = None
     Nitems = 0
-    shape = tuple()
+    shape = tuple()  # noqa
```
steven-murray: what's wrong with doing this?

pyuvsim/tests/test_simsetup.py
```diff
@@ -550,13 +529,22 @@ def test_param_select_redundant():
...
 def check_uvdata_keyword_init(case):
-    base_kwargs = dict(
-        antenna_layout_filepath=os.path.join(SIM_DATA_PATH, "test_config/triangle_bl_layout.csv"),
-        telescope_location=(-30.72152777777791, 21.428305555555557, 1073.0000000093132),
-        telescope_name="HERA", Nfreqs=10, start_freq=1e8, bandwidth=1e8, Ntimes=60,
-        integration_time=100.0, start_time=2458101.0, polarization_array=['xx'],
-        no_autos=True, write_files=False, run_check=True
-    )
+    base_kwargs = {
+        "antenna_layout_filepath": os.path.join(SIM_DATA_PATH,
```
steven-murray: this line will be flagged by black if we ever do that.

pyuvsim/tests/test_simsetup.py
```diff
@@ -807,7 +793,9 @@ def test_multi_analytic_beams():
 
     telescope_location = (-30.72152777777791, 21.428305555555557, 1073.0000000093132)
     telescope_name = 'SKA'
-    beam_specs = {0: 'airy, diameter=14', 1: 'airy, diameter=20', 2: 'gaussian, sigma=0.5'}
+    beam_specs = {0: {'type': 'airy', 'diameter': 14},
```
steven-murray: also would be flagged by black (need newline)

pyuvsim/version.py
```diff
@@ -54,7 +54,9 @@ def construct_version_info():
 
     try:
         git_origin = _get_git_output(['config', '--get', 'remote.origin.url'], capture_stderr=True)
-        if git_origin.split('/')[-1] != 'pyuvsim.git':  # pragma: no cover
+        print('git origin:', git_origin)
```
steven-murray: I thought we updated to using `pygitversion` in pyuvsim? Or has that not been merged yet?

setup.py
```diff
@@ -36,7 +36,7 @@
     'scripts': glob.glob('scripts/*'),
     'version': version.version,
     'include_package_data': True,
-    'install_requires': reqs,
+    'install_requires': ['numpy>=1.15', 'scipy', 'astropy>=4.0', 'pyyaml', 'pyuvdata'],
```
steven-murray: If we're gonna have the `requirements.txt` file (which I am agnostic to), then it's probably better just to load that file in here (as it used to be). I know it's less explicit, but it's pretty easy to see where it's being read from, and the file is easily human-readable. Otherwise, we could just remove the requirements.txt as well.

.gitignore
bhazelton: yeah, I need to read more about that. This is the existing structure, but the folder wasn't excluded via the gitignore (it was deleted at the end of the test session, but I noticed the files in my git status while tests were running so added it to the ignore for good measure).

.travis.yml
bhazelton: we could, my intention is to remove Travis as soon as the rest of it is working, so I went with the minimal change here.

ci/min_deps.yaml
bhazelton: we could. This is how we do it other places, mainly I was wanting to enforce that tests are skipped properly so that you can run tests if you don't install any extras.

pyuvsim/mpi.py
bhazelton: This is flagged as unnecessary by the `flake8-comprehensions` plugin, which wants me to use `shape = ()` to initialize a tuple. I found that less readable so added the `# noqa` to stop it complaining at me.

pyuvsim/tests/test_simsetup.py
bhazelton: There are a number of black changes that should happen, but I'm inclined to wait for a break in the PR traffic for that. I did make an issue about it though.

pyuvsim/version.py
bhazelton: We don't, yet, nor is there an outstanding PR to do so. I actually started to change it in this PR and then decided to wait for a different for a new one to limit the number of things I changed in this PR.

setup.py
bhazelton: I will just remove the requirements.txt file, I find that approach less readable.

ci/min_deps.yaml
bhazelton: So it was minimal packages rather than minimal versions I was targeting. But we could do the latter as well.

setup.py
aelanman: I am also in favor of removing the requirements.txt, but we should keep two things in mind:
-The requirements.txt file was a request from the JOSS reviewer a year ago. I guess he preferred installing with pip?
-There is a line in the README.md (under installation) that references it.

bhazelton: What the JOSS reviewer wanted was for the installation to be automatic with pip. We have achieved that (except for pyradiosky) with another approach so I don't think the requirements file is really needed. We should push on a pip package for pyradiosky so that we can make it easier to install pyuvsim. Thanks for pointing out the required readme change.

bhazelton: Ok, I removed it and updated the readme and setup.py. I also added instructions for installing pyradiosky, which can't be done with just the `pip install` step. Please take a look and let me know if I've missed anything.

bhazelton:
Should I remove the Travis builds in this PR?

setup.py
bhazelton: Just realized I forgot about installing the pre-commit hook, just added that as well.

.gitignore
steven-murray: I'll create an issue for it.

.travis.yml
steven-murray: Ah yeah, fine.

pyuvsim/mpi.py
steven-murray: ech. FWIW I find `shape = ()` perfectly fine (and less annoying that havine a `# noqa` appended). But this is fine.

pyuvsim/version.py
steven-murray: Great. Maybe I'm thinking of `pyuvdata` haha. I get so confused.

README.md
```diff
@@ -28,47 +30,68 @@ Key elements of this approach include:
...
-    $ conda activate pyuvsim
-    $ conda env update -n pyuvsim -f environment.yml
-    $ pip install -e .
-
-This will install extra dependencies required for testing/development as well as the
-standard ones.
-
-The second-to-last line may also be replaced by `pip install -r requirements.txt` if you
-do not care about using `conda`.
+following packages before installing `pyuvsim`:
```
steven-murray: Hmmm, so we've traded having this explicitly listed in `requirements.txt` for having them explicitly listed here? This still seems redundant. Can we just point people to the `setup.py` here?

bhazelton: To be fair, I didn't add them here, they were already here. I just turned them into a bulleted list rather than an inline list. It can be hard to maintain separate lists, but I also think clarity in the readme is helpful.

steven-murray: Ah, fair. If you're happy to have the maintenance overhead, then that's fine.

ci/min_deps.yaml
steven-murray: Ahhhh, I see. I do think an actual minimum version dep test might be a good idea (perhaps not on every PR?) but we can do that another time.

steven-murray:
I'd remove the travis tests -- they're not reporting because Travis has now changed everything over to using the GitHub Apps API through travis.com rather than travis.org and it can no longer communicate. I had the same thing happen in another repo. You can fix it, but if you're losing travis anyway, might as well just ditch it now.

PR #273, opened 2020-03-07
Benchmarking

aelanman:
## Description

Provides some tools for running benchmarking simulations on a SLURM-enabled computing cluster and tracking the results. This is useful for ensuring that future changes to core functions don't drastically reduce performance. These changes provide scripts to submit a job with controlled resources and scale, track the memory usage and runtime, and record this information for posterity. There is also an (experimental) script that can read the line profiler data and attempt to estimate how the runtime scales.

## Motivation and Context

There has been much discussion about ways to ensure that future changes to `pyuvsim` don't cause unexpected slowdowns.

## Types of changes

- [ ] Version change

## Checklist:

New feature checklist:
(There is a README in the benchmarking directory to describe how to use these tools).


aelanman:
After a great deal of digging, I found that the the line_profiler function `enable_by_count` is causing coverage issues. Any lines run by tests _after_ the profiler is enabled will not be reported as covered by `pytest-cov`. I've raised an issue on the line_profiler repo (https://github.com/rkern/line_profiler/issues/179#issue-587085050), though I may port it over to pytest-cov instead.

In the meantime, I've streamlined the profiler functions as best I can and marked some lines as not covered. I can confirm manually that all of the lines marked with "pragma: nocover" _are_ actually run by the tests.

benchmarking/benchmark.py
```diff
@@ -0,0 +1,250 @@
...
+import yaml
+import os
+import h5py
+from pyradiosky import write_healpix_hdf5
+
+from pyuvsim.simsetup import _write_layout_csv, freq_array_to_params
+
+
+def settings_setup(settings_file, outdir=None):
+    """ Parse settings file and make other variables."""
```
bhazelton: Please flesh out the docstrings in this module with parameters and returns sections.

benchmarking/BENCHMARKS.log
```diff
@@ -0,0 +1,2 @@
+Date/Time                 	uvsim_version	SettingsFile 	cpus-per-task	Ntasks	Nnodes	MemLimit	Ntimes	Nbls	Nfreqs	Nsrcs	Nsrcs_part	MaxRSS [GiB]	Runtime
+2020-03-06-21:40:47.584612	1.1.2.       	settings.yaml	2            	10    	1     	50G     	21    	500 	256   	3072 	1         	13.071      	4:32:16.373965
```
bhazelton: Is this all the info we need to tell if there are speed changes? I guess I'm worried about whether different lines could be wildly different if something about the hardware is different. Maybe just adding the hostname would help with tracking that?

pyuvsim/tests/conftest.py
```diff
@@ -42,6 +42,8 @@ def setup_and_teardown_package():
 
 
 @pytest.fixture(autouse=True)
-def ignore_center_deprecation():
+def ignore_deprecation():
     warnings.filterwarnings('ignore', message='The default for the `center` keyword has changed.',
                             category=DeprecationWarning)
```
bhazelton: This warning was removed in version 1.5 of pyuvdata (we're now on 2.0), so I think you can take this filter out.

pyuvsim/profiling.py
```diff
@@ -69,13 +103,12 @@ def set_profiler(func_list=default_profile_funcs, rank=0, outfile_name='time_pro
...
+            outfile_raw_name = outfile_prefix + ".lprof"
             atexit.register(prof.dump_stats, outfile_raw_name)
-
+        setattr(prof, 'rank', rank)     # Add "rank" as an attribute to the profiler.
+        setattr(prof, 'meta_file', outfile_prefix + '_meta.out')
         prof.enable_by_count()
 
 
-def get_profiler():
+def get_profiler():  # pragma: nocover
```
bhazelton: Let's add a comment to the lines you are leaving out of coverage because of the interaction between line_profiler and the coverage calculator to the effect that they are in fact covered by tests and that the `# pragma: nocover` can be removed when that bug is fixed.

pyuvsim/uvsim.py
```diff
@@ -403,6 +404,28 @@ def run_uvdata_uvsim(input_uv, beam_list, beam_dict=None, catalog=None):
     if rank == 0:
         print("Calculations Complete.")
 
+    # If profiling is active, save meta data:
+    from .profiling import prof     # noqa
+    if hasattr(prof, 'meta_file'):  # pragma: nocover
```
bhazelton: Let's add a comment to the lines you are leaving out of coverage because of the interaction between line_profiler and the coverage calculator to the effect that they are in fact covered by tests and that the # pragma: nocover can be removed when that bug is fixed.

benchmarking/benchmark.py
aelanman: I've added parameters and returns to the docstrings.

benchmarking/BENCHMARKS.log
aelanman: I've added the host machine name and url, which should be enough to indicate where the job was run. It's just one more column. I'd like to keep this file readable, so I'm hesitant to add too many more columns.

pyuvsim/tests/conftest.py
aelanman: I've removed the filter. Glad this is fixed!

pyuvsim/profiling.py
aelanman: I've left comments here and in uvsim to explain the lack of coverage for now, which also reference the relevant issue on the line_profiler github page.

pyuvsim/uvsim.py
aelanman: as above

benchmarking/BENCHMARKS.log
bhazelton: that sounds good, thanks!

PR #282, opened 2020-04-03
Reduce recalculations

### LINKED ISSUE #237: Make beam calculations not repeated for each baseline if beams are the same
Hopefully a fast speed up.

aelanman:
## Description

The addition of the frequency axis to sky models introduced a lot of repeated calculations. The `local_coherency` matrix was being recalculated (at all frequencies and source positions) for every time, frequency, and baseline, when it only needs to be recalculated once per time. The Jones matrices only need to be calculated when the time, frequency, or beam type changes, and can be stored otherwise. These changes further reduce the repetition of these calculations.

There are corresponding changes to `pyradiosky` that enable further speedups, but this branch is compatible with the current master branch on `pyradiosky`. A quick test using the benchmarking tools shows a speedup from 3.5 hours to about 10 minutes with both of these branches in place (on `pyradiosky` and `pyuyvsim`), with no  significant impact on memory usage.

## Motivation and Context

Speed is a limiting factor for the usefulness of `pyuvsim`. This offers a considerable speedup for the majority of use cases.

Closes #237 

## Types of changes

- [ ] Version change

## Checklist:

New feature checklist:

### DIFF
--- a/benchmarking/BENCHMARKS.log
+++ b/benchmarking/BENCHMARKS.log
@@ -1,2 +1,3 @@
 Date/Time                 	uvsim_version	HostName                   	SettingsFile 	cpus-per-task	Ntasks	Nnodes	MemLimit	Ntimes	Nbls	Nfreqs	Nsrcs	Nsrcs_part	MaxRSS [GiB]	Runtime       
 2020-03-06-21:40:47.584612	1.1.2.       	node474.oscar.ccv.brown.edu	settings.yaml	2            	10    	1     	50G     	21    	500 	256   	3072 	1         	13.071      	4:32:16.373965
+2020-03-26-17:54:43.959585	1.1.2.       	node474.oscar.ccv.brown.edu	settings.yaml	2            	10    	1     	50G     	21    	500 	256   	3072 	1         	12.187      	3:24:00.635442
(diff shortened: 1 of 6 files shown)


benchmarking/BENCHMARKS.log
```diff
@@ -1,2 +1,3 @@
 Date/Time                 	uvsim_version	HostName                   	SettingsFile 	cpus-per-task	Ntasks	Nnodes	MemLimit	Ntimes	Nbls	Nfreqs	Nsrcs	Nsrcs_part	MaxRSS [GiB]	Runtime       
-2020-03-06-21:40:47.584612	1.1.2.       	node474.oscar.ccv.brown.edu	settings.yaml	2            	10    	1     	50G     	21    	500 	256   	3072 	1         	13.071      	4:32:16.373965
\ No newline at end of file
+2020-03-06-21:40:47.584612	1.1.2.       	node474.oscar.ccv.brown.edu	settings.yaml	2            	10    	1     	50G     	21    	500 	256   	3072 	1         	13.071      	4:32:16.373965
+2020-03-26-17:54:43.959585	1.1.2.       	node474.oscar.ccv.brown.edu	settings.yaml	2            	10    	1     	50G     	21    	500 	256   	3072 	1         	12.187      	3:24:00.635442
```
bhazelton, line 3: Is this last line supposed to be 20x faster than the previous one?

aelanman: Whoops, I forgot to commit the log change.

PR #284, opened 2020-04-14
Beam upgrades

aelanman:
## Description

Support for setting spline interpolation order in `UVBeam._inter_az_za_rect_spline` from telescope param file.

Also enables keeping UVBeam data array in a shared memory window to avoid memory bloat.

## Motivation and Context

Enables more control over spatial interpolation in UVBeam.
Avoids a significant memory bloat issue when using UVBeams.

## Types of changes

- [ ] Version change

New feature checklist:

### DIFF
--- a/pyuvsim/antenna.py
+++ b/pyuvsim/antenna.py
@@ -80,12 +80,20 @@ class Antenna(object):
         spline_opts = None
         if isinstance(array.beam_list, BeamList):
             spline_opts = array.beam_list.spline_interp_opts
-        interp_data, interp_basis_vector = \
-            array.beam_list[self.beam_id].interp(az_array=source_az,
-                                                 za_array=source_za,
-                                                 freq_array=freq,
-                                                 reuse_spline=reuse_spline,
-                                                 spline_opts=spline_opts)
+
+        interp_kwargs = {'az_array' : source_az, 'za_array' : source_za,
+                         'freq_array' : freq, 'reuse_spline' : reuse_spline,
+                         'spline_opts' : spline_opts}
+
+        # For backwards compatibility.
+        try:
+            interp_data, interp_basis_vector = \
+                array.beam_list[self.beam_id].interp(**interp_kwargs)
+        except TypeError:
+            interp_kwargs.pop('spline_opts')
+            interp_data, interp_basis_vector = \
+                array.beam_list[self.beam_id].interp(**interp_kwargs)
+
         Ncomponents = source_za.shape[-1]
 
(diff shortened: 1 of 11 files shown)


pyuvsim/antenna.py
```diff
@@ -73,13 +74,26 @@ def get_beam_jones(self, array, source_alt_az, frequency, reuse_spline=True,
...
+                         'spline_opts' : spline_opts}
+
+        # For backwards compatibility.
+        try:
+            interp_data, interp_basis_vector = \
+                array.beam_list[self.beam_id].interp(**interp_kwargs)
+        except TypeError:
+            interp_kwargs.pop('spline_opts')
+            interp_data, interp_basis_vector = \
+                array.beam_list[self.beam_id].interp(**interp_kwargs)
```
bhazelton, line 95: Is this for old versions of pyuvdata? This will result in the wrong thing being run. I think it should error with a helpful message about needing a newer version of pyuvdata if you want to use spline options.

pyuvsim/simsetup.py
```diff
@@ -358,6 +358,8 @@ def initialize_catalog_from_params(obs_params, input_uv=None):
 
 def _construct_beam_list(beam_ids, telconfig):
     beam_list = BeamList([])
+    if 'spline_interp_opts' in telconfig.keys():
```
bhazelton, line 361: should we add anything to the docs about this option?

pyuvsim/antenna.py
aelanman: You're right. I've changed it to error when the interp call fails. However, to cover the except block we'd have to add a new test case using the older pyuvdata version. I've set the lines to be ignored by codecov.

pyuvsim/simsetup.py
aelanman: I've added this and `freq_interp_kind` to the documentation on parameter files.

aelanman:
I tried to add a test that actually confirms that the scipy is using the correct spline order, but cannot get anything returned from get_beam_jones that reflects the order of splines used. With the UVBeam model available, the difference between linear and fifth order splines is not noticeable.

I can confirm, however, that scipy does error if you try to give it zeroth order splines, which are not allowed, so it does look like the spline order is correctly passed to the interpolation function. _However_ this error is raised from wrapped C code, not as a Python exception, and so cannot be caught by pytest.

Overall -- I can't write a test that makes sure of it, but from my interactions with the new feature I am confident that the parameters are being used correctly.

PR #285, opened 2020-04-17
Fixes for SkyModel deprecations

### LINKED ISSUE #278: Switch to using SkyModel.read_gleam_catalog
When (this PR)[https://github.com/RadioAstronomySoftwareGroup/pyradiosky/pull/49] on pyradiosky is merged, SkyModel will have a function specifically for reading GLEAM and a more general function for reading VOTables (with deafults equivalent to reading GLEAM). We should switch here to using `read_gleam_catalog` so that the `read_votable_catalog` defaults can be set more generally.

bhazelton:
## Description

Fix code that was causing deprecation warnings from pyradiosky, update votable handling.

## Motivation and Context

closes #278

## Types of changes

- [ ] Version change

## Checklist:

Bug fix checklist:


pyuvsim/data/test_config/param_1time_1src_testvot.yaml
```diff
@@ -7,6 +7,11 @@ freq:
   channel_width: 1000000.0
 sources:
   catalog: '../single_source.vot'
+  table_name: single
```
aelanman, line 10: We should probably have it in the documentation that these additional fields are necessary.

bhazelton: I updated the docs.

PR #292, opened 2020-05-05
Split source bugfix

### LINKED ISSUE #291: Bug with UVBeam simulations
When using UVBeam, there is an error on the einsum step in `apply_beam`:

```
ValueError: operands could not be broadcast together with remapped shapes [original->remapped]: (2,2,85448)->(2,newaxis,85448,2,newaxis) (2,2,57935)->(57935,2,2) (2,2,85448)->(2,85448,newaxis,2)
```
I suspect that it's just that the coherency matrix isn't updating before the jones matrices, so the horizon mask applied to the jones matrix calculation doesn't match what the local coherency has. If this is the case, it's a simple fix (move one line earlier).

aelanman:
## Description

This fixes a major oversight with the recent speedup changes. The speedup works by only updating positions and evaluating jones matrices / coherencies when they need to be updated, and also only computing jones matrices/coherencies for sources above the horizon. This is achieved by keeping certain quantities on the UVEngine for reuse and only reevaluating them when time/frequency/beam type, etc. changes.

The oversight was that I forgot that the source array itself gets split, so those same quantities need to update when a new source chunk is processed. This PR fixes that, making everything update when the source array changes.

## Motivation and Context

A large sky model will be processed in chunks on each MPI process, but source positions will not be properly updated when the "chunk" changes. This caused an error in einsum in `apply_beam`.

closes #291 

## Types of changes

- [ ] Version change

## Checklist:

Bug fix checklist:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,6 +13,7 @@
 - The definition of the Airy beam now uses the exact value of c, not 3e8.
 
 ### Fixed
+- Ensure that source positions, coherency matrices, and Jones matrices are updated at the right times.
 - Error early if the task list is too long for gather
 
 ## [1.1.2] - 2020-2-14
--- a/docs/example_configs/bl_lite_mixed.yaml
+++ b/docs/example_configs/bl_lite_mixed.yaml
@@ -1,11 +1,11 @@
 beam_paths:
   0: hera.uvbeam
   1:
-    - type: airy
-      diameter: 16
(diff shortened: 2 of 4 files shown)


aelanman:
I was hesitant to add to the changelog, since these changes fix a bug introduced by other unreleased changes. I wasn't sure if adding this to the 'Fixed' section suggests that the problem was introduced in a prior release.

PR #297, opened 2020-05-15
lunarsky support

### LINKED ISSUE #5: coherency_calc currently assumes earth-based observatories
to generalize we need to split this function into array orientation and actual coherency calculation

aelanman:
## Description

Provides compatibility with the `lunarsky` module, if installed. If the keyword "world: moon" is in the telescope config yaml file, the `telescope_location` will be interpreted as a lat/lon/height position for an observatory on the Moon. The output UVData object has `extra_keyword['world'] = moon` to identify when this was done.

There are also a few docstring updates, and the addition of a "quiet" option to suppress stdout printing.

## Motivation and Context

It has been a long-standing issue to allow pyuvsim to simulate observations from the lunar surface for the FARSIDE experiment. The `lunarsky` module provides an extension to `astropy`'s coordinate classes, using the JPL SPICE toolkit to calculate transformations to and from lunar reference frames. `pyradiosky` already support `lunarsky`. These changes extend support to `pyuvsim`.

closes #5 

## Types of changes

- [ ] Version change

## Checklist:

New feature checklist:

### DIFF
--- a/ci/tests.yaml
+++ b/ci/tests.yaml
@@ -19,4 +19,4 @@ dependencies:
   - setuptools_scm
   - pip:
     - pyradiosky
-    - git+https://github.com/aelanman/lunarsky.git
+    - lunarsky
--- a/ci/tests.yaml
+++ b/ci/tests.yaml
@@ -19,3 +19,4 @@ dependencies:
   - setuptools_scm
   - pip:
     - pyradiosky
+    - git+https://github.com/aelanman/lunarsky.git
--- a/pyuvsim/tests/test_simsetup.py
+++ b/pyuvsim/tests/test_simsetup.py
@@ -9,7 +9,6 @@ import shutil
 import numpy as np
 import pytest
 import yaml
-import warnings
 from astropy import units
 from astropy.coordinates import Angle, SkyCoord, EarthLocation
 from pyuvdata import UVBeam, UVData
@@ -1002,6 +1001,9 @@ def test_moon_lsts():
     assert new_obj.check()
 
 
+@pytest.mark.filterwarnings("ignore:The _ra parameters are not")
+@pytest.mark.filterwarnings("ignore:The _dec parameters are not")
+@pytest.mark.filterwarnings("ignore:Future equality does not pass")
 def test_mock_catalog_moon():
     # A mock catalog made with a MoonLocation.
     pytest.importorskip('lunarsky')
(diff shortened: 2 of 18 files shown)


ci/tests.yaml
```diff
@@ -19,3 +19,4 @@ dependencies:
   - setuptools_scm
   - pip:
     - pyradiosky
+    - git+https://github.com/aelanman/lunarsky.git
```
bhazelton, line 22: It would be good to get lunarsky on pypi sometime soon (not required for this PR).

pyuvsim/tests/test_simsetup.py
```diff
@@ -966,3 +968,56 @@ def test_beamopts_init():
...
+    eloc = EarthLocation.from_geodetic(24.433, 0.6875)
+    emock, ekwds = pyuvsim.simsetup.create_mock_catalog(time, 'hera_text', array_location=eloc)
+
+    assert mkwds['world'] == 'moon'
+    assert ekwds['world'] == 'earth'
+
+    # Simple check that the given lat/lon were interpreted differently in each call.
+    with warnings.catch_warnings():
+        warnings.simplefilter('ignore')
+        assert mmock != emock
```
bhazelton, line 1023: does pytest.warn not work here?

ci/tests.yaml
aelanman: I added lunarsky to PyPi. No reason not to at this point.

pyuvsim/tests/test_simsetup.py
aelanman: I'm not trying to test the warnings that are made. The expectation is that the source positions aren't close together, which means that SkyModel was warning that the ra/dec aren't close to the future tolerance. I've replaced it with a pytest warning filter.

ci/tests.yaml
bhazelton: great!

ci/tests.yaml
```diff
@@ -19,3 +19,4 @@ dependencies:
   - setuptools_scm
   - pip:
     - pyradiosky
+    - lunarsky
```
bhazelton, line 22: we should probably add this to the `environment.yaml` file in the top level of the repo as well. And we should add it to the extras_require options `all` and `dev` and consider making a new option `moon`.

aelanman: Thanks for the suggestion! I've added it to the environment.yaml, and the extras_require options, and added a `moon` option.

PR #300, opened 2020-05-22
Big MPI operations

### LINKED ISSUE #289: Break up large MPI operations
As a follow up to issue #288 : Instead of breaking, pyuvsim should be able to split up gather/broadcast operations when they are too large for MPI to do in one go.

aelanman:
## Description

Adds the functions `big_bcast` and `big_gather` to `mpi.py`, which allow for broadcasting and gathering data in volumes larger than the 2 GiB limit in MPI.

## Motivation and Context

MPI operations are limited to handling data smaller than about 2 GiB, because it uses 32-bit integers to describe the total size of broadcast or gathered objects. Several packages, such as [BigMPI](https://www.mcs.anl.gov/papers/P5210-1014.pdf) exist to help avoid this limit, but either aren't Python-compatible or aren't stable enough to make a dependency.

These changes avoid the limit by splitting broadcasts and gathers into multiple parts if they exceed the limit. This may be a little slower, but jobs won't crash broadcasting large beams or catalogs, or gathering large numbers of tasks.

It isn't really possible to make a test that attempts a very large MPI operation, but this has been tested with a simulation that was breaking before.

closes #289 

## Types of changes

- [ ] Version change

## Checklist:

Bug fix checklist:

### DIFF
--- a/pyuvsim/mpi.py
+++ b/pyuvsim/mpi.py
@@ -21,8 +21,9 @@ world_comm = None
 node_comm = None
 rank_comm = None
 
-# Maximum size of objects that can be handled by a single MPI operation.
-INT_MAX = 2**32 - 1
+
+# Split serialized objects into chunks of 2 GiB
+INT_MAX = 2**31 - 1
 
 
 def set_mpi_excepthook(mpi_comm):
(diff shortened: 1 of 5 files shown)


aelanman:
The uncovered lines won't be reached without a unit test with multiple MPI processes.

aelanman:
The uncovered lines shouldn't require any large tests, but they're only called from rank != 0 processes, so until multiprocess tests are in place they can't be reached. I think they should be covered some day, but that's true of many `pragma: nocover` lines.

There's an open issue to add MPI-enabled tests. I think I'll just add a comment to that issue that whenever that's done, the non-covered lines should be re-examined.

bhazelton:
Ok, sounds good.

PR #303, opened 2020-06-03
SkyModelData

### LINKED ISSUE #293: Use SkyModel methods rather than functions.
A recent PR on pyradiosky moved most of the functions to methods (with deprecation). We should fix the calls in pyuvsim to use the new methods.

aelanman:
## Description

Replaces the use of recarrays for sharing SkyModel data with a new object, defined in `pyuvsim.simsetup`, that handles all the required behavior. Also cuts down on memory bloat due to repeating the `freq_array` Ncomponents times when storing as a column in a recarray.

## Motivation and Context

pyradiosky inherited the "recarray" form of sky models from `pyuvsim`, which was a hacky solution to share sky model data with all MPI processes in a container that is MPI-friendly. It also made sense with how votables and text files were initially read in (as numpy arrays). This partially saved on memory, because much of the data could be put in shared memory windows on each node. As pyradiosky evolved, however, the corresponding recarray definition got more complicated. The addition of the `freq_array` for instance, required that the recarray carry around a column that's just repetitions of the `freq_array`, which is horribly inefficient.

The main impediment to directly using SkyModels in MPI is that several attributes are defined as astropy Quantities (or derived classes, such as Angles, Latitudes, Longitudes), and this information is lost when sharing the data in MPI, unless the slow serialization methods are used (as in `mpi4py.MPI.bcast`). I've added a function to allow broadcasting astropy Quantities to shared memory, but haven't used it much.

This defines a new class, `SkyModelData`, which carries the SkyModel parameters as ndarrays. These arrays can be shared across MPI processes by running the SkyModelData.share() method, which puts arrays in shared memory (if that array has an axis of length `Ncomponents`).

Further, it minimizes Stokes information that is broadcast by only keeping non-stokes I flux values for components which have nonzero polarization. Most simulations won't have many polarized sources, so this can reduce required memory.

closes #293 

## Types of changes

- [ ] Version change

## Checklist:

New feature checklist:

### DIFF
--- a/pyuvsim/simsetup.py
+++ b/pyuvsim/simsetup.py
@@ -639,8 +639,9 @@ def initialize_catalog_from_params(obs_params, input_uv=None, return_recarray=Tr
 
     # Make SkyModelData to share it.
     if return_recarray:
-        warnings.warn("initialize_catalog_from_params will not "
-                      "return recarray by default in the future.", PendingDeprecationWarning)
+        warnings.warn("initialize_catalog_from_params will return a SkyModel instance, not "
+                      "a recarray, by default in the future. Set keyword "
+                      "`return_recarray=True` to ensure recarray is returned.", PendingDeprecationWarning)
         sky = sky.to_recarray()
 
     return sky, source_list_name
--- a/pyuvsim/simsetup.py
+++ b/pyuvsim/simsetup.py
@@ -331,6 +331,7 @@ class SkyModelData:
     polarized = None
     nside = None
     hpx_inds = None
+    flux_unit = None
 
     put_in_shared = ['stokes_I', 'stokes_Q', 'stokes_U', 'stokes_V', 'polarized',
                      'ra', 'dec', 'reference_frequency', 'spectral_index', 'hpx_inds']
@@ -350,7 +351,13 @@ class SkyModelData:
             stokes_in = sky_in.stokes
 
             if isinstance(stokes_in, units.Quantity):
-                stokes_in = stokes_in.to('Jy').value
+                if stokes_in.unit.is_equivalent("Jy"):
+                    stokes_in = stokes_in.to("Jy").value
+                    self.flux_unit = "Jy"
+                elif stokes_in.unit.is_equivalent("K"):
+                    stokes_in = stokes_in.to("K").value
+                    self.flux_unit = "K"
+
             self.stokes_I = stokes_in[0, ...]
 
             if sky_in._n_polarized > 0:
@@ -409,7 +416,6 @@ class SkyModelData:
         if self.freq_array is not None:
             new_sky.freq_array = self.freq_array
         if self.hpx_inds is not None:
-            print("hello", flush=True)
             new_sky.hpx_inds = self.hpx_inds[inds]
 
         if self.polarized is not None:
@@ -476,8 +482,8 @@ class SkyModelData:
             stokes_use[2, :, self.polarized] = self.stokes_U.T
             stokes_use[3, :, self.polarized] = self.stokes_V.T
 
-        if units.Quantity in pyradiosky.SkyModel()._stokes.expected_type:
-            stokes_use *= units.Jy
+        if self.flux_unit is not None:
+            stokes_use *= units.Unit(self.flux_unit)
 
         other = {}
         if self.spectral_type in ['full', 'subband']:
--- a/pyuvsim/simsetup.py
+++ b/pyuvsim/simsetup.py
@@ -329,8 +329,8 @@ class SkyModelData:
     spectral_index = None
     polarized = None
 
-    ncomp_attrs = ['stokes_I', 'stokes_Q', 'stokes_U', 'stokes_V',
-                   'ra', 'dec', 'reference_frequency', 'spectral_index', 'name']
+    put_in_shared = ['stokes_I', 'stokes_Q', 'stokes_U', 'stokes_V', 'polarized',
+                     'ra', 'dec', 'reference_frequency', 'spectral_index', 'name']
 
     def __init__(self, sky_in=None):
         # Collect relevant attributes.
@@ -343,8 +343,8 @@ class SkyModelData:
             self.Nfreqs = sky_in.Nfreqs
             self.stokes_I = sky_in.stokes[0, ...]
 
-            self.polarized = sky_in._polarized
             if sky_in._n_polarized > 0:
+                self.polarized = sky_in._polarized
                 Q, U, V = sky_in.stokes[1:, :, sky_in._polarized]
                 self.stokes_Q = Q
                 self.stokes_U = U
@@ -357,13 +357,15 @@ class SkyModelData:
             if sky_in._spectral_index.required:
                 self.spectral_index = sky_in.spectral_index
 
-    def __getitem__(self, inds):
+    def subselect(self, inds):
         """
-        Subselect, returning a new SkyModelData object
+        Subselect, returning a new SkyModelData object.
+
+        To avoid copying data, the indices must be a range object.
 
         Parameters
         ----------
-        inds: slice or index array
+        inds: range or index array
             Indices to select along the component axis.
 
         Returns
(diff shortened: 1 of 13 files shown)


pyuvsim/simsetup.py
```diff
@@ -285,29 +287,199 @@ def create_mock_catalog(time, arrangement='zenith', array_location=None, Nsrcs=N
...
+                self.stokes_V = V
+
+            if sky_in._freq_array.required:
+                self.freq_array = sky_in.freq_array.to("Hz").value
+            if sky_in._reference_frequency.required:
+                self.reference_frequency = sky_in.reference_frequency.to("Hz").value
+            if sky_in._spectral_index.required:
+                self.spectral_index = sky_in.spectral_index
+
+    def __getitem__(self, inds):
```
bhazelton, line 360: After reading some about `__getitem__` this seems to be a non-standard implementation -- it's returning a downselected copy of the object which is not what `__getitem__` is generally used for.

pyuvsim/simsetup.py
```diff
@@ -285,29 +287,199 @@ def create_mock_catalog(time, arrangement='zenith', array_location=None, Nsrcs=N
...
+    stokes_Q = None
+    stokes_U = None
+    stokes_V = None
+    freq_array = None
+    reference_frequency = None
+    spectral_index = None
+    polarized = None
+
+    ncomp_attrs = ['stokes_I', 'stokes_Q', 'stokes_U', 'stokes_V',
+                   'ra', 'dec', 'reference_frequency', 'spectral_index', 'name']
```
bhazelton, line 333: Is this supposed to list attributes that have length `Ncomponents`? I don't think stokes_Q/U/V generally do have that length, do they? I think that they have the length of the number of polarized sources. If that's the case, I'm concerned about where you use it in `__getitem__` to downselect using the component indices. On the other hand, you have a test that looks like it is testing that this works, so I'm confused.

If stokes_Q/U/V do have length `Ncomponents`, I don't see why to use this new object -- it will have the same size as the SkyModel object.

aelanman: Using a new class helps because it keeps the data in numpy arrays with fixed units instead of in Angles/Latitudes/Quantity etc. It's a lot easier to handle with MPI this way.

You're right though, about the stokes_Q/U/V arrays... they don't have length `Ncomponents` unless every source is polarized. This does break __getitem__ so I'll have to fix that too. Thanks for pointing it out!

pyuvsim/uvsim.py
```diff
@@ -273,17 +273,17 @@ def uvdata_to_task_iter(task_ids, input_uv, catalog, beam_list, beam_dict, Nsky_
         raise TypeError("input_uv must be UVData object")
 
     #   Skymodel will now be passed in as a catalog array.
-    if not isinstance(catalog, np.ndarray):
+    if not isinstance(catalog, SkyModelData):
         raise TypeError("catalog must be a record array")
```
bhazelton, line 277: I think this error message is wrong.

aelanman: Thanks, it's fixed now.

aelanman:
@bhazelton I've renamed the `__getitem__` function to `subselect`, and changed its syntax to properly handle the polarization arrays. There's some other changes in there too to ensure that, under normal operation, the arrays are not copied (i.e., views are returned). I have a test in place to make sure that the arrays aren't copied, but to be safe I'm running a simulation with a large catalog to make sure it doesn't use more memory than previously.

The subselect method is, unfortunately, necessary for the (albeit, rare) case that Npus > Nbls * Ntimes * Nfreqs, but Nsrcs > Npus. In that case, the source axis is split across MPI processes, instead of in chunks on each rank.

pyuvsim/simsetup.py
```diff
@@ -369,24 +574,24 @@ def initialize_catalog_from_params(obs_params, input_uv=None):
...
+        sky, mock_keywords = create_mock_catalog(time, **mock_keywords)
         mock_keyvals = [str(key) + str(val) for key, val in mock_keywords.items()]
         source_list_name = 'mock_' + "_".join(mock_keyvals)
     elif isinstance(catalog, str):
         source_list_name = os.path.basename(catalog)
         if not os.path.isfile(catalog):
             catalog = os.path.join(param_dict['config_path'], catalog)
         if catalog.endswith("txt"):
-            catalog = pyradiosky.read_text_catalog(catalog, return_table=True)
+            sky = pyradiosky.read_text_catalog(catalog)
```
bhazelton, line 585: The function you're calling here is deprecated. Can you use `SkyModel.read_text_catalog`?

pyuvsim/simsetup.py
```diff
@@ -369,24 +574,24 @@ def initialize_catalog_from_params(obs_params, input_uv=None):
...
             if 'gleam' in catalog:
                 if "spectral_type" in source_params:
                     spectral_type = source_params["spectral_type"]
                 else:
                     warnings.warn("No spectral_type specified for GLEAM, using 'flat'.")
                     spectral_type = "flat"
-                catalog = pyradiosky.skymodel.read_gleam_catalog(
-                    catalog, spectral_type=spectral_type, return_table=True
+                sky = pyradiosky.skymodel.read_gleam_catalog(
+                    catalog, spectral_type=spectral_type
```
bhazelton, line 594: The function you're calling here is deprecated. Can you use `SkyModel.read_gleam_catalog`?

pyuvsim/simsetup.py
```diff
@@ -409,20 +614,23 @@ def initialize_catalog_from_params(obs_params, input_uv=None):
                     vo_params["dec_column"] = source_params["dec_column"]
 
                 vo_params["reference_frequency"] = None
-                catalog = pyradiosky.read_votable_catalog(
-                    catalog, **vo_params, return_table=True
+                sky = pyradiosky.read_votable_catalog(
+                    catalog, **vo_params
```
bhazelton, line 618: The function you're calling here is deprecated. Can you use `SkyModel.read_votable_catalog`?

pyuvsim/simsetup.py
```diff
@@ -409,20 +614,23 @@ def initialize_catalog_from_params(obs_params, input_uv=None):
...
                 )
         elif catalog.endswith('hdf5'):
             hpmap, inds, freqs = pyradiosky.read_healpix_hdf5(catalog)
             sky = pyradiosky.healpix_to_sky(hpmap, inds, freqs)
```
bhazelton, line 622: These functions are deprecated. Can you use `SkyModel.read_healpix_hdf5`?

pyuvsim/simsetup.py
aelanman: updated!

aelanman:
@bhazelton I've replaced the deprecated function calls, but I'd like to wait on this until we've finished making the stokes array a Quantity in pyradiosky.

aelanman:
@bhazelton This should be ready now. It's compatible with the new changes to pyradiosky and with the current release.

I have another branch which won't be compatible with the current release, so if possible we should make a new pyradiosky release soon.

aelanman:
Never mind... I just realized that `hera_sim` is relying on `simsetup.initialize_catalog_from_params` to return a recarray, so I'll need to add backwards compatibility.

We should probably add a github action to run the `hera_sim` tests, since it relies on pyuvsim so heavily. I'm not really sure how to do that though.

steven-murray:
@aelanman if there's a better way of using the `pyuvsim.simsetup` routines now (or directly through `pyradiosky`, I'd be happy to implement it in `hera_sim`!

aelanman:
@steven-murray The main function (`initialize_catalog_from_params`) hasn't changed much, except to return a `SkyModel` directly instead of a recarray. I've restored the option to return a recarray and made it the default for compatibility, but ultimately we don't want that to be the default behavior.

It made some sense for `initialize_catalog...` to return a recarray back when that was an intermediate step to reading in catalogs (astropy's VOTable reader and the numpy text parser both returned recarrays). That has been changed in pyradiosky, so the function was making a SkyModel and then converting it to recarray (presumably then to be converted back by the user...). This was only necessary for MPI, but this branch fully removes pyuvsim's need to convert catalogs to recarrays.

PR #310, opened 2020-07-07
MPI-enabled tests

### LINKED ISSUE #126: Make unit tests that use multiple MPI ranks
More and more I'm finding bugs that only come up when running a simulation with more than one rank. Although we test the MPI framework, we don't actually do any unit tests with multiple processing units. I believe it is possible to run nosetests with mpirun, but it would probably require adjusting how we call them in travis.

Adding to the First release milestone, because I don't think we can really trust our code coverage unless this is achieved.

### LINKED ISSUE #235: Counter test failing intermittently
`test_mpi_uvsim.py::test_mpi_counter` fails about 1% of the time in python 3. This has only started coming up recently.

The MPI counter works by running a thread on the rank 0 process that monitors for messages telling it to increment. It's possible that pytest doesn't work well with an embedded thread, or the counter doesn't work so well when messages are sent at it too quickly (as is happening in the test). So far I'm unable to spot what's going wrong.

This is a minor utility that's only used for tracking simulation progress and reporting it to the user, so it's not a huge problem.

aelanman:
## Description

Defines a new pytest marker that tells pytest to run a given test as a subprocess. Within the subprocess, the given test is run using `mpirun` and a user-specified number of MPI processes (up to 10). This marker is applied to several of the tests in `test_mpi.py`. Using this, I found that adding a win.Fence() to the counter made the counter test pass consistently, so it will no longer be skipped.

A command line option `--nompi` will disable the parallelized tests.

I'm not sure how to test it automatically, but I've included some hooks that will print the stack traces separately for each mpi process, labelled by COMM_WORLD rank, when a test fails. This can be seen by temporarily breaking a parallelized test. There is also a settable "timeout" option to handle cases where MPI causes a deadlock.

(This was branched off of reorganize_tests, and so should be merged there or merged into master after PR #309 is merged)

## Motivation and Context

It has been a long-standing issue to parallelize our tests, since many bugs have only come up when running pyuvsim on multiple MPI ranks.

Closes #235 
Closes #126 

## Types of changes

- [ ] Version change

## Checklist:

Bug fix checklist:

New feature checklist:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,7 @@
 ## [1.2.0] - 2020-7-20
 
 ### Added
+- Support for unit tests parallelized with MPI.
 - Diffuse models in mock catalogs, via the analytic_diffuse module.
 - quantity_shared_bcast function to allow objects derived from astropy.units.Quantity to use shared memory broadcasting.
 - SkyModelData class to replace recarray conversion in pyradiosky.
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,9 +1,14 @@
 # Changelog
 
-## [1.2.0] - 2020-7-20
+## [Unreleased]
 
 ### Added
 - Support for unit tests parallelized with MPI.
+
+
+## [1.2.0] - 2020-7-20
+
+### Added
 - Diffuse models in mock catalogs, via the analytic_diffuse module.
 - quantity_shared_bcast function to allow objects derived from astropy.units.Quantity to use shared memory broadcasting.
 - SkyModelData class to replace recarray conversion in pyradiosky.
--- a/pyuvsim/tests/conftest.py
+++ b/pyuvsim/tests/conftest.py
@@ -5,6 +5,7 @@
 """Testing environment setup and teardown for pytest."""
 import os
 import warnings
+from subprocess import check_call
 
 import pytest
 from astropy.time import Time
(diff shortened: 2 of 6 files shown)


CHANGELOG.md
```diff
@@ -3,6 +3,7 @@
 ## [1.2.0] - 2020-7-20
 
 ### Added
+- Support for unit tests parallelized with MPI.
```
bhazelton, line 6: This needs to be moved up to under unreleased changes.

pyuvsim/tests/conftest.py
```diff
@@ -24,7 +37,138 @@ def pytest_collection_modifyitems(session, config, items):
...
+
+
+def pytest_addoption(parser):
+    parser.addoption("--nompi", action="store_true", help="skip mpi-parallelized tests.")
+
+
+def pytest_runtest_setup(item):
+    if 'parallel' in item.keywords:
+        if mpi is None:
+            pytest.skip("Need mpi4py to run parallelized test.")
```
bhazelton, line 59: test -> tests (presumably there is more than one?)

aelanman:
Thanks for the review @bhazelton . Do you know of anyone else who might review the pytest hooks?

bhazelton:
@mkolopanis can you look at this?

aelanman:
Here's an example -- I modified `test_mpi_funcs` to raise an assertion error on rank 0 and a value error on rank 1.

error_demo

Another example -- I inserted the following into uvsim.py, which causes MPI to stall:
```
if rank == 1:
    var=1
    comm.gather(var, root=0)
```
(`gather` needs to be called from all processes at once). I then ran `test_run.py::test_run_paramfile_uvsim`. This prints a much longer stack trace, but indicates that the test failed because of a timeout (which I set to 10 seconds in this case, but is 70 sec by default):
timeout_error_demo

This isn't as neatly formatted, but the relevant information is there. An MPI hang doesn't raise an exception, so I don't know of a way to identify programmatically where it happens, but a timeout can potentially identify if it happens and ensure that other tests still run.

aelanman:
@mkolopanis I've commented here with a couple examples of the error output if you'd like to take a look.

mkolopanis:
Ah yes thank you. I guess my only other question is do we want it to print the rank that gave each error?

aelanman:
I'd think so. The ranks of failing processes are printed in the first case (in cyan). When the subprocess times out, however, it's usually because there isn't a single failing process, but a stalled MPI state.

mkolopanis:
Oh I see. I missed that one. Thanks.

aelanman:
Thanks Matt!

PR #321, opened 2020-08-04
Default pixel interp

### LINKED ISSUE #316: healpix format beams support in telescope config file
The default angular interpolation is `az_za_rect_spline `, which does not work with beamfiles described in healpix. Specifying `interpolation_function: 'healpix_simple'` does not seem to work either since we can not pass the filename in dictionary mode, while the interpolation function needs to be passed to each beam ID.

aelanman:
## Description

Fixes a small bug in `Antenna.get_beam_jones`. It was defaulting the `interpolation_method` on the UVBeam to `az_za_simple` (from the keywords), which was overwriting whatever interpolation function was set on the UVBeams. This resets the default to None.

This also shifts responsibility on the user calling the function to ensure that either `interpolation_method` is set on the UVBeam object, or that it's passed in as a keyword argument.

## Motivation and Context

This is required for pyuvsim to support UVBeams with healpix maps for their pixel data.

Potentially fixes #316 

## Types of changes

- [ ] Version change

## Checklist:

Bug fix checklist:

### DIFF
--- a/pyuvsim/antenna.py
+++ b/pyuvsim/antenna.py
@@ -26,7 +26,7 @@ class Antenna(object):
         self.beam_id = beam_id
 
     def get_beam_jones(self, array, source_alt_az, frequency, reuse_spline=True,
-                       interpolation_function='az_za_simple', freq_interp_kind=None):
+                       interpolation_function=None, freq_interp_kind=None):
         """
         2x2 array of Efield vectors in Az/Alt
 
@@ -45,19 +45,16 @@ class Antenna(object):
         interpolation_function: str
             Set the angular interpolation function on the UVBeam.
             See UVBeam.interp for options.
-            Defaults to az_za_simple.
         freq_interp_kind : str
             Interpolation method for frequencies.
             Note -- This overrides whatever method may be set on the
             UVBeam objects.
         Returns
         -------
-
         jones_matrix : (2,2) ndarray, dtype float
             The first axis is feed, the second axis is vector component
             on the sky in az/za.
         """
-
         # get_direction_jones needs to be defined on UVBeam
         # 2x2 array of Efield vectors in alt/az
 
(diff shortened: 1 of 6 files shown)


bhazelton:
I'm a little worried this is going to make it so that previous config files don't work because they didn't specify the interpolation function. Can we default it to az/za but still make it so that if it is set in the config file we don't ignore that setting?

aelanman:
@bhazelton Thanks for the feedback. I've added a new line that sets the interpolation_function to az_za_simple if the UVBeam.interpolation_function is still None after potentially being set by keyword. Ideally, the interpolation_function should be defined before this point is reached. Maybe I should add a warning?

bhazelton:
I think a warning is a good idea.

pyuvsim/antenna.py
```diff
@@ -71,14 +69,23 @@ def get_beam_jones(self, array, source_alt_az, frequency, reuse_spline=True,
...
-            array.beam_list[self.beam_id].freq_interp_kind = freq_interp_kind
+            beam.freq_interp_kind = freq_interp_kind
 
         if interpolation_function is not None:
-            array.beam_list[self.beam_id].interpolation_function = interpolation_function
+            beam.interpolation_function = interpolation_function
+
+        # If interpolation_function is unset, or the beam lacks the attribute
+        # (i.e., is analytic) then default to az_za_simple.
+        if getattr(beam, 'interpolation_function', 'analytic') is None:
```
bhazelton, line 85: I find the comment a little confusing -- it doesn't default to anything for an analytic beam if I'm reading the code right. Which is fine, there's no interpolation to do for an analytic beam.

aelanman: I've rewritten the lines. I just wanted something that would assign a default to UVBeams when the interpolation_function isn't set but not do anything to analytic beams, and this was quick way to do it.

PR #329, opened 2020-12-04
Distribute random source uniformly by solid angle

### LINKED ISSUE #328: Draw random point sources uniformly distributed over solid angle.
https://github.com/RadioAstronomySoftwareGroup/pyuvsim/blob/d571b4847b1bf884236f13994a2d79f429c8483a/pyuvsim/simsetup.py#L249

random sources are currently drawn distributed uniformly in azimuth and altitude but to get a uniform distribution per solid angle, one should actually do something like this

https://mathworld.wolfram.com/SpherePointPicking.html

aelanman:
## Description

Modifies the random distribution mock catalog to ensure sources are distributed uniformly by solid angle.

## Motivation and Context

The existing implementation of random sources chooses uniformly-distributed altitudes and azimuths, which leads to a nonuniform distribution of points on the sphere (clustering near the poles). To ensure the random points are truly distributed uniformly on the sphere, the distribution of altitude had to be adjusted.

Closes #328 

## Types of changes

- [ ] Version change

## Checklist:

For all pull requests:

Bug fix checklist:


bhazelton:
Was this random distribution used in any of our reference sims?

aelanman:
It's not in any ref sims.

pyuvsim/tests/test_simsetup.py
```diff
@@ -884,48 +884,91 @@ def test_uvfits_to_config():
...
+    [
+        ('cross', 'mock_cross_2458098.27471.txt'),
+        ('hera_text', 'mock_hera_text_2458098.27471.txt'),
+        ('long-line', 'mock_long-line_2458098.27471.txt'),
+        ('off-zenith', 'mock_off-zenith_2458098.27471.txt'),
+        ('triangle', 'mock_triangle_2458098.27471.txt'),
+        ('random', 'mock_random_2458098.27471.txt'),
+        ('zenith', 'mock_zenith_2458098.27471.txt'),
+    ]
+)
```
bhazelton, line 898: nice use of parametrize!

pyuvsim/tests/test_simsetup.py
```diff
@@ -884,48 +884,91 @@ def test_uvfits_to_config():
...
+    hp = astropy_healpix.HEALPix(nside)
+
+    inds = hp.lonlat_to_healpix(Longitude(az, unit='rad'), Latitude(alt, unit='rad'))
+    un, counts = np.unique(inds, return_counts=True)
+
+    # counts should be Poisson-distributed with rate lambda = nsrcs / npix
+    # variance and mean should be close to lambda
+    lam = Nsrcs / npix
+    assert np.isclose(np.mean(counts), lam, atol=1.0)
+    assert np.isclose(np.var(counts), lam, atol=1.0)
```
bhazelton, line 965: This is a clever way to handle testing a random distribution. The only thing I worry about slightly with random things in tests are flaky tests that fail randomly. Do you have a sense for how likely this is to fail with these tolerances?

bhazelton (approved):
Looks pretty good to me, as long as the test for the random distribution is unlikely to fail in a flaky way.

pyuvsim/tests/test_simsetup.py
aelanman: The random generator is given the same seed for each run, so it should always turn up the same points. That said, I just ran it 500 times without the fixed rseed, so it's actually random, and didn't have any failures.

pyuvsim/tests/test_simsetup.py
aelanman: Thanks :)

pyuvsim/tests/test_simsetup.py
bhazelton: Ah, I didn't catch that it was seeded. I'm good with that, thanks.

PR #331, opened 2021-01-19
Fix CI and bugs in our standard yaml file

### LINKED ISSUE #330: Failed Developer Installation through environment.yml
Running `conda env create -f environment.yml` will return 

```
PackagesNotFoundError: The following packages are not available from current channels:
  - pyradiosky
```
I found it was solved by deleting the line `- pyradiosky` under `dependencies`. And `pyradiosky ` is actually installed via 
```
- pip:
    - pyradiosky>=0.1.0
```

bhazelton:
## Description

- Stop pushing every build to testpypi, fix all github workflows to use the `setup-miniconda` action.
- Start RTD builds on PRs
- Fix bugs in our standard conda yaml file for developers.

## Motivation and Context

Fixes #330 

## Types of changes

- [ ] Version change

## Checklist:

For all pull requests:


environment.yml
```diff
@@ -9,9 +9,9 @@ dependencies:
   - mpi4py>=3.0.0
   - numpy>=1.15
   - pip
+  - pre-commit
```
mkolopanis, line 12: I'm a little weary about this choice. I have pulled `pre-commit` from conda before and gotten a broken build. But I'm not one to dictate where to get packages from. Maybe just make a mental note in case things get weird one day.

mkolopanis (approved):
This looks pretty good to me. I'm a bit confused by the codecov report stating that coverage is gone since none of these changes should affect that. Also an admin will need to turn off the python 3.6 requirement on checks before tit can be merged.

.github/workflows/testsuite.yaml
```diff
@@ -84,34 +97,43 @@ jobs:
...
         run: |
-          source activate $ENV_NAME
           python -m pytest --cov=pyuvsim --cov-config=.coveragerc --cov-report xml:./coverage.xml --junitxml=test-reports/xunit.xml
 
       - uses: codecov/codecov-action@master
-        if: success()
+        if: matrix.os == 'ubuntu-latest' && success()
         with:
           token: ${{secrets.CODECOV_TOKEN}} #required
           file: ./coverage.xml #optional
```
mkolopanis, line 139: also another little note, we can also upload coverage for other platforms by doing this
```
        name: Upload Coverage
        if: success()
        run: |
          bash <(curl -s https://codecov.io/bash) -t ${{ secrets.CODECOV_TOKEN }}
```

environment.yml
bhazelton: weary or wary? ;) I guess I just prefer to get everything I can from conda and this is what I've been doing on all my local environments. But I don't want to push people to something that's likely to be broken. I could add a comment within the yaml if you think that's a good idea.

.github/workflows/testsuite.yaml
bhazelton: I was actually just trying to make it match the full test matrix above, but I just realized we don't need it here because this is only built on linux. I'm not sure we need to upload the coverage from Mac builds, but I guess we could.

bhazelton:
I don't understand the codecov issues, digging into that now.

environment.yml
mkolopanis: Clearly I'm too weary to be wary about the differences between those words. The dangers of not consuming coffee in the morning. If it is working now, maybe I just got a single bad build. I think if this works, let's just stick with it.

.github/workflows/testsuite.yaml
mkolopanis: Good point. Just an option, and written down too, if we want to do it later.

bhazelton: I actually think that the codecov issues are related to this, all the coverage drops are related to running without all the dependencies, so I think the min deps upload didn't happen somehow. I'll take it out. I'm happy to add the Mac coverage, it shouldn't really take any extra time, so no harm.

bhazelton: Actually, according to the readme on that action, it should work on Mac & Windows: 
https://github.com/codecov/codecov-action

mkolopanis: Oh that is great news! My knowledge is probably from older versions.

bhazelton: I'll give it a shot...

mkolopanis (approved):
Good finds on that coverage.

PR #337, opened 2021-02-22
Fix reading catalog files via pyradiosky

bhazelton:
## Description

The current handling for reading in catalog files via pyradiosky can fail silently because of a missing `else` clause. It also does not handle our new standard `skyh5` file formats.

I also made the version requirements for pyuvdata & pyradiosky consistent throughout the repo. This was prompted by an error in the RTD build.

## Motivation and Context

This fixes a bug reported in Slack by Bharat.

## Types of changes

- [ ] Version change

## Checklist:

For all pull requests:

Bug fix checklist:

### DIFF
--- a/pyuvsim/simsetup.py
+++ b/pyuvsim/simsetup.py
@@ -697,7 +697,9 @@ def initialize_catalog_from_params(obs_params, input_uv=None, return_recarray=Tr
                     vo_params[param] = source_params[param]
                 for param, default in warn_params.items():
                     if param not in source_params:
-                        warnings.warn(f"No {param} specified for {catalog}, using default.")
+                        warnings.warn(
+                            f"No {param} specified for {catalog}, using default: {default}."
+                        )
                         vo_params[param] = default
                     else:
                         vo_params[param] = source_params[param]
--- a/pyuvsim/simsetup.py
+++ b/pyuvsim/simsetup.py
@@ -674,6 +674,7 @@ def initialize_catalog_from_params(obs_params, input_uv=None, return_recarray=Tr
         sky = pyradiosky.SkyModel()
         if not os.path.isfile(catalog):
             catalog = os.path.join(param_dict['config_path'], catalog)
+
         if catalog.endswith("txt"):
             sky.read_text_catalog(catalog)
         elif catalog.endswith('vot'):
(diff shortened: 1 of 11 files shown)


pyuvsim/simsetup.py
```diff
@@ -688,30 +689,35 @@ def initialize_catalog_from_params(obs_params, input_uv=None, return_recarray=Tr
...
-                    vo_params["dec_column"] = source_params["dec_column"]
+                required_params = ["table_name", "id_column", "flux_columns"]
+                warn_params = {"ra_column": "RAJ2000", "dec_column": "DEJ2000"}
+                for param in required_params:
+                    if param not in source_params:
+                        raise ValueError(f"No {param} specified for {catalog}.")
+                    vo_params[param] = source_params[param]
+                for param, default in warn_params.items():
+                    if param not in source_params:
+                        warnings.warn(f"No {param} specified for {catalog}, using default.")
```
mkolopanis, line 700: should we print out what the defualt is here too?

pyuvsim/simsetup.py
```diff
@@ -688,30 +689,35 @@ def initialize_catalog_from_params(obs_params, input_uv=None, return_recarray=Tr
...
+        elif catalog.endswith('skyh5'):
+            sky.read_skyh5(catalog)
         elif catalog.endswith('hdf5'):
-            sky.read_healpix_hdf5(catalog)
+            # In principal, this could either be a skyh5 or an old-style healvis hdf5 file
+            # Try skyh5 first, if that doesn't work, use the old method
+            try:
+                sky.read_skyh5(catalog)
+            except ValueError:
+                sky.read_healpix_hdf5(catalog)
```
mkolopanis, line 715: Should this _also_ be wrapped in a try/except or are we just banking on pyradiosky to raise from here if it is not a healpix hdf5 file?

mkolopanis (commented):
Super simple Here. I think I have a few questions but other than that looking good.

pyuvsim/simsetup.py
bhazelton: sure, I can add that. It's a column name.

pyuvsim/simsetup.py
bhazelton: pyradiosky checks if it's not right and raises a useful error.

mkolopanis: okay that sounds good enough for me then.

mkolopanis (approved):
Looks good to me. I don't know what why coverage is unhappy now though.

bhazelton:
The coverage report makes no sense. It says that `pyuvsim/__init__.py` is a new file, which it definitely is not. I don't know what's going on with codecov today, it's not making sense across several repos.

PR #342, opened 2021-03-11
Parallel test bug resolution

### LINKED ISSUE #341: Bug with parallelized tests which fail in serial mode
MPI-enabled tests work by spawning a subprocess to run pytest with `mpirun`. The pytest hook `pytest_runtest_call` identifies if a given test has the `parallel` marker and spawns the subprocess. However, if the subprocess finishes without error (that is, the test passes), the main thread then goes on to run the test (without parallelization, in serial mode). If you have a test that will only pass with multiple MPI processes, it will fail at this step. This hasn't come up before because all of our tests have passed in serial mode if they pass in parallel.

There's a fairly simple fix to this -- If the subprocess completes (i.e., the parallelized test passes), replace the test function in the main thread with a blank function that doesn't do anything. This has been implemented here:
https://github.com/RadioAstronomySoftwareGroup/pyuvsim/blob/4724f417999025a4080d753549a28cd7c08caac8/pyuvsim/tests/conftest.py#L113-L116

aelanman:
## Description

Generalizes parallel tests to run with different "root" processes. Writing this uncovered a bug in parallel pytest, which is also resolved here.
Also adds a `ParallelFlag` class which isn't used, but might be useful. This provides a set of flags across all MPI processes which can be checked asynchronously by the root process.

## Motivation and Context

Currently, parallelized tests which pass with Npus > 1 but fail with Npus == 1 will fail in general with pytest due to this bug. Ideally, the `parallel` marker means that a given test should be pass with the specified Npus, and whether or not it passes with Npus==1 is irrelevant.

Fixes #341 

## Types of changes

- [ ] Version change

## Checklist:

For all pull requests:
- [ ] I have read the [contribution guide](CONTRIBUTING.md).
- [ ] My code follows the code style of this project.

Bug fix checklist:

New feature checklist:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,7 @@
 - Use remote memory access to collect finished visibility data, without serialization.
 
 ### Fixed
+- Bug MPI-enabled tests which caused failure on tests that wouldn't pass in serial mode.
 - Fix bugs in reading catalogs using pyradiosky for `skyh5` files and files with unusual extension names.
 - Corrects the distribution of random points for the random mock catalog.
 - pixel interpolation was defaulting to az_za_simple for all beams, breaking healpix-coord UVBeams.
--- a/pyuvsim/mpi.py
+++ b/pyuvsim/mpi.py
@@ -386,52 +386,6 @@ class Counter:
         return nval[0]
 
 
-class ParallelFlag:
-    """
-    Uses RMA to monitor status on multiple processes.
-
-    A boolean array of length comm.size, which each process
-    can flip without blocking.
-    """
-
-    def __init__(self, comm=None, base_rank=0):
-        self.base_rank = base_rank
-        if comm is None:
-            comm = world_comm.Dup()
-        self.comm = comm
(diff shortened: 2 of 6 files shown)


bhazelton (dismissed):
This looks fine to me, but I'm really not an MPI expert. I'll approve for now, but won't merge yet until @mkolopanis can weigh in if he wants to.

mkolopanis (commented):
Thanks for finding and fixing that test execution bug. I suppose my only real comment is why add this ParallelFlag class if we're not using it for anything? Feels like extra code we're just carrying around at that point. I agree it could be potentially useful but its implementation is not terribly complex either and I think we could easily re-create it if we find a need.

aelanman:
@mkolopanis I just didn't want to delete it after writing it, but it doesn't need to be there. I'll remove it.

mkolopanis:
I totally understand that. Never fun to write something, then throw it away.

mkolopanis (approved):
Looks good to me! Thanks @aelanman

PR #351, opened 2021-06-16
Propagate Clobber to skyh5 and uvh5

### LINKED ISSUE #340: Propagate clobber to uvh5 writer
this line should have the clobber keyword propagated into it.
https://github.com/RadioAstronomySoftwareGroup/pyuvsim/blob/d701412698bac3674ea3ccc3d4318feb34e5aa1e/pyuvsim/utils.py#L231

mkolopanis:
fixes #340 
propagates the clobber keyword from pyradiosky into some tests that call `write_skyh5` requires RadioAstronomySoftwareGroup/pyradiosky#144

### DIFF
--- a/pyuvsim/utils.py
+++ b/pyuvsim/utils.py
@@ -181,7 +181,6 @@ def write_uvdata(
     return_filename=False,
     dryrun=False,
     out_format=None,
-    clobber=True,
 ):
     """
     Parse output file information from parameters and write uvfits to file.
--- a/pyuvsim/utils.py
+++ b/pyuvsim/utils.py
@@ -175,19 +175,32 @@ def check_file_exists_and_increment(filepath, extension=None):
     return filepath
 
 
-def write_uvdata(uv_obj, param_dict, return_filename=False, dryrun=False, out_format=None):
+def write_uvdata(
+    uv_obj,
+    param_dict,
+    return_filename=False,
+    dryrun=False,
+    out_format=None,
+    clobber=True,
+):
     """
     Parse output file information from parameters and write uvfits to file.
 
-    Args:
-        uv_obj: UVData object to write out.
-        param_dict: parameter dictionary defining output path, filename, and
-                    whether or not to clobber.
-        return_filename: (Default false) Return the file path
-        dryrun: (Default false) Don't write to file.
-        out_format: (Default uvfits) Write as uvfits/miriad/uvh5
+    Parameters
+    ----------
+    uv_obj : UVData Object
+        The object to be written out.
+    param_dict : Dict
+        parameter dictionary defining output path, filename, and whether or not to clobber.
+    return_filename : Bool
+        (Default false) Return the file path
+    dryrun : Bool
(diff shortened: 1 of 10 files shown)


pyuvsim/utils.py
```diff
@@ -175,19 +175,32 @@ def check_file_exists_and_increment(filepath, extension=None):
...
+    uv_obj : UVData Object
+        The object to be written out.
+    param_dict : Dict
+        parameter dictionary defining output path, filename, and whether or not to clobber.
+    return_filename : Bool
+        (Default false) Return the file path
+    dryrun : Bool
+        (Default false) Don't write to file.
+    out_format : Str
+        (Default uvfits) Write as uvfits/miriad/uvh5
```
bhazelton, line 200: I'm confused -- should a user control file clobbering using the `clobber` keyword or using the param_dict? The clobber keyword is not listed in the docstring Parameters section, but the description of the param_dict suggests that's the way to control it.

bhazelton (commented):
I'm confused by the code vs test changes -- the test changes seem to be all related to SkyModel.write_skyh5 while the code changes seem to be all about UVData.write_uvh5

pyuvsim/utils.py
mkolopanis: I have no idea honestly. Looks like it is picked out of the param dict in line 238. So presumably somewhere along the way the user gets to control it?

mkolopanis:
I think I ended up smashing two things together in my head for this because of the word "clobber". Clobber was added to skyh5 writer and needed to be updated in the tests here (since there's some on the fly catalog writing? ) but also that the uvh5 writer did not propagate whatever `clobber` keyword was being sent to the UVData writer.

pyuvsim/utils.py
bhazelton: But you added a `clobber` keyword to the function as well that is not documented in the docstring. Does it work in addition to the param_dict?

mkolopanis: oh! that's a mistake! thanks for catching. It is part of the param dict. Looks like I missed that, added it, then found it as part of the param dict and never removed that line

pyuvsim/utils.py
```diff
@@ -175,19 +175,32 @@ def check_file_exists_and_increment(filepath, extension=None):
...
 
 
-def write_uvdata(uv_obj, param_dict, return_filename=False, dryrun=False, out_format=None):
+def write_uvdata(
+    uv_obj,
+    param_dict,
+    return_filename=False,
+    dryrun=False,
+    out_format=None,
+    clobber=True,
```
mkolopanis, line 184:
```suggestion
```

bhazelton:
I think it'd be good to have at least one test that checks that the clobber keyword is passed to the UVData write method(s) properly, especially since it's going through the param_dict mechanism.

mkolopanis:
Sounds good I'll get that done. Sorry for the confusion with doing two things!

bhazelton (approved):
Looks good, thanks!

PR #352, opened 2021-06-18
re-enable codecov uploads

mkolopanis:
pin codecov action v1.5.2 until codecov/codecov-action#372 is resolved


jpober (approved):
hera_sim failures are expected.  It's not clear why readthedocs is still pending -- looks like it finished and passed when you check the readthedocs page itself?  Sure, why not!

steven-murray:
There is now a hera_sim v1 on master -- it has a few new dependencies, which it looks like aren't being installed for this test. If you want to get hera_sim back up and running, might be good to make sure these install.

mkolopanis:
thanks for the update, we're using #350 to get the hera_sim external dependency going again :D

PR #357, opened 2021-07-22
fix: SkyModelData.name must be an array not a list

steven-murray:
## Description
This does two things:

1. Adds `psutil` as a dependency (rather than optional extra). `psutil` is used in a default run, so it should not be optional.
2. Makes sure the `SkyModel.name` attribute is an array, not a list. By default, the `SkyModel` object creates a list for a healpix map. But then doing `subselect` breaks because an index range can't be used on lists.

## Types of changes

- [ ] Version change

## Checklist:

For all pull requests:

Bug fix checklist:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,7 +14,7 @@
 - Fix bugs in reading catalogs using pyradiosky for `skyh5` files and files with unusual extension names.
 - Corrects the distribution of random points for the random mock catalog.
 - pixel interpolation was defaulting to az_za_simple for all beams, breaking healpix-coord UVBeams.
-
+- `SkyModel.name` is now coerced to an array.
 ## [1.2.0] - 2020-7-20
 
 ### Added
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,11 +10,12 @@
 - Use remote memory access to collect finished visibility data, without serialization.
 
 ### Fixed
+- `SkyModel.name` is now coerced to an array.
 - Bug MPI-enabled tests which caused failure on tests that wouldn't pass in serial mode.
 - Fix bugs in reading catalogs using pyradiosky for `skyh5` files and files with unusual extension names.
 - Corrects the distribution of random points for the random mock catalog.
 - pixel interpolation was defaulting to az_za_simple for all beams, breaking healpix-coord UVBeams.
-- `SkyModel.name` is now coerced to an array.
+
 ## [1.2.0] - 2020-7-20
 
 ### Added
--- a/pyuvsim/simsetup.py
+++ b/pyuvsim/simsetup.py
@@ -418,7 +418,10 @@ class SkyModelData:
     def __init__(self, sky_in=None):
         # Collect relevant attributes.
         if sky_in is not None:
-            self.name = np.asarray(sky_in.name) if sky_in.name is not None else None
+            if sky_in.name is not None:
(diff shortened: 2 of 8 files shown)


CHANGELOG.md
```diff
@@ -14,7 +14,7 @@
 - Fix bugs in reading catalogs using pyradiosky for `skyh5` files and files with unusual extension names.
 - Corrects the distribution of random points for the random mock catalog.
 - pixel interpolation was defaulting to az_za_simple for all beams, breaking healpix-coord UVBeams.
-
+- `SkyModel.name` is now coerced to an array.
```
bhazelton, line 17: Please add this at the top of its section so its in reverse chronological order.

pyuvsim/simsetup.py
```diff
@@ -418,7 +418,7 @@ class SkyModelData:
     def __init__(self, sky_in=None):
         # Collect relevant attributes.
         if sky_in is not None:
-            self.name = sky_in.name
+            self.name = np.asarray(sky_in.name) if sky_in.name is not None else None
```
bhazelton, line 421: I don't love one line if/else statements, I find them hard to read.

bhazelton (commented):
Thanks for fixing these @steven-murray ! Can you also update the README to move psutil to the required rather than optional dependency list? It also appears that it's not in the `ci/min_deps.yaml` file, I'm not sure why the min_deps tests are passing without it, but it should be added to that yaml as well as the rtd_env and publish yamls in the `ci` folder (mostly so we aren't confused in the future about why it's not there).

pyuvsim/utils.py
```diff
@@ -11,7 +11,7 @@
 try:
     import psutil
     HAVE_PSUTIL = True
-except ImportError:
+except ImportError:  # pragma: no cover
     HAVE_PSUTIL = False
```
bhazelton, line 15: Oh, this whole block should be replaced with just `import psutil`. If it's a required dependency we don't need to do this dance.

steven-murray: Oh yeah, good catch!

CHANGELOG.md
steven-murray: Ahhh, I'll never remember which way round that goes. Done.

pyuvsim/simsetup.py
steven-murray: oh, I love them when they're small like this, but that's OK. Easy fix!

steven-murray:
OK, I did all the things you requested, @bhazelton. Ball has been knocked back into your court :-)

pyuvsim/utils.py
bhazelton: It doesn't look like you made this change yet.

steven-murray: huh. I swear I did. Maybe I didn't CTRL+S

steven-murray: OK it's done now.

bhazelton:
Hmm, everything is failing, I think there's an indent error. Do you have the pre-commit hook installed?

steven-murray:
OK. All good now. Sorry for my daftness.

bhazelton:
Ack, I just merged another PR and in the process I created a merge conflict with this one. Sorry! I can fix it if you'd like or you can.

bhazelton (approved):
This looks good to me now, thanks!

bhazelton (approved):

PR #358, opened 2021-08-03
Multi progbar backend

mkolopanis:
Adds `tqdm` progress bar as a backend option for `run_uvdata_uvsim` also implements a send and receive protocol for run_uvdata_uvsim where the tasks are chunked and sent out to PUs on request.

## Motivation and Context

I like how tqdm looks honestly and was tired of two gross branches continually rebasing. This allows for all backends depending on the user's preference. I think I would prefer having some kind of plugin or something to allow custom progress bars but that sounds much more complicated. 

TODO:
- [ ] Show reference sims are identical
- [ ] benchmarking reference sims
- [ ] Documentation/tutorial about backend selection
- [ ] changelog

## Types of changes

- [ ] Version change

## Checklist:

For all pull requests:

New feature checklist:


pyuvsim/uvsim.py
```diff
@@ -360,7 +360,251 @@ def _check_ntasks_valid(Ntasks_tot):
...
+                if tag == mpi.tags.READY:
+                    pass
+                elif tag == mpi.tags.DONE:
+                    uv_container.data_array[msg[1]] += msg[0]
+                    pbar.update(1)
+
+                elif tag == mpi.tags.EXIT:
+                    completed_workers += 1
+                else:
+                    raise ValueError(f"{msg} {tag}")
```
mkolopanis, line 478: There's two lines in here that just _won't_ ever get touched by tests. the `READY` tag is never sent back from a worker thread, and I think the else catch was a debug thing that just never went away just in case. Ideally the "enum" defined in the mpi module is an exhaustive list of statuses that can be sent back. Would it be preferred to just take out these lines?

aelanman:
While I fully support switching to tqdm over progsteps, I'm still not convinced of the push/pull model for task distribution. Has it been shown conclusively that active distribution performs better than the existing system? It looks like it should be capable of about the same performance except with an added overhead from using blocking sends to return visibilities to rank 0.

mkolopanis:
Definitely part of this endeavor is to just get the technology in the repo right alongside the other things in order to test it better. There were branches before, but it was almost impossible to rebase or something after a new feature was brought in. Even just sitting on this branch, it is much easier to make a comparison. 

I think one thing we should do is to get a few really good timing tests. We could always drop the send/recv part in the end.

PR #360, opened 2021-09-06
feat: ability to do consistency checks on beam lists

steven-murray:
## Description

This adds the `check_consistency` method on the `BeamList` class, which can be manually called to check that all of the consistuent beams have some consistent properties. It also adds a test for both passing and failing cases.

## Motivation and Context

In simulation, often you're assuming that all your beams come from the same telescope, and have some set of similar properties (eg, all are efield/power and all have the same set of polarizations/feeds etc). This is not *always* the case, or at least, more general simulators will allow different beams to have different properties and have ways of simulating each one independently. However, many simulators will want some way to check consistency before going off and running. This adds  a simple method that can be called in such a case.

I am not the expert on what people may or may not want to check for. I added a keyword `check_pols` to check for polarization-related stuff (which is almost everything) because I don't know if all simulators will want to check for that. There are other UVBeam parameters that I ignored because I don't know enough about them, but perhaps should be added to the checks (eg. Ncomponents_vec and frequency-related parameters). 

I also opted to raise an error if one of the beams doesn't *have* a certain parameter but the rest do, as well as if they both have the parameter but they're different. I'm not sure this is always the right thing to do. For instance, AnalyticBeam objects don't define `Nfeeds` even though they're `efield` beams by default. Not sure how that works.

## Types of changes

- [ ] Version change

## Checklist:

For all pull requests:

New feature checklist:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,7 @@
 ## [1.2.1] - 2021-10-13
 
 ### Added
+- New `check_consistency` method for `BeamList` objects.
 - Support for writing out measurement set files.
 - Support for unit tests parallelized with MPI.
 - Require that future changes not drastically increase runtime for current capabilities.
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,7 @@
 ## [Unreleased]
 
 ### Added
+- New `check_consistency` method for `BeamList` objects.
 - Ability to simulate UVData objects where Nblts != Nbls * Ntimes. Currently only supported with direct `run_uvdata_uvsim` calls.
 
 ### Fixed
@@ -11,7 +12,6 @@
 ## [1.2.1] - 2021-10-13
 
 ### Added
-- New `check_consistency` method for `BeamList` objects.
 - Support for writing out measurement set files.
 - Support for unit tests parallelized with MPI.
 - Require that future changes not drastically increase runtime for current capabilities.
--- a/pyuvsim/telescope.py
+++ b/pyuvsim/telescope.py
@@ -11,6 +11,12 @@ from .analyticbeam import AnalyticBeam
 from . import mpi
 
 
+class BeamConsistencyError(Exception):
+    """An Exception class to mark inconsistencies among beams in a BeamList."""
+
+    pass
+
+
 class BeamList:
     """
     A container for the set of beam models and related parameters.
@@ -44,19 +50,17 @@ class BeamList:
         Passing in a mixture of strings and objects will error.
     uvb_params : dict (optional)
         Options to set uvb_params, overriding settings from passed-in UVBeam objects.
-
-    check
+    check : bool
         Whether to perform a consistency check on the beams (i.e. asserting that several
         of their defining parameters are the same for all beams in the list).
-
-    force_check
+    force_check : bool
         The consistency check is only possible for object-beams (not string mode). If
         `force_check` is True, it will convert beams to object-mode to force the check
         to run (then convert back to string mode).
 
     Raises
     ------
-    ValueError :
+    ValueError
         For an invalid beam_list (mix of strings and objects).
 
     Notes
@@ -406,8 +410,7 @@ class BeamList:
 
     @classmethod
     def _obj_to_str(cls, beam_model):
-        # Convert beam objects to strings that may generate them.
-
+        """Convert beam objects to strings that may generate them."""
         if isinstance(beam_model, str):
             return beam_model
         if isinstance(beam_model, AnalyticBeam):
@@ -486,12 +489,6 @@ class BeamList:
 
         if check:
             self.check_consistency()
-
-
-class BeamConsistencyError(Exception):
-    """An Exception class to mark inconsistencies among beams in a BeamList."""
-
-    pass
 
 
 class Telescope:
--- a/pyuvsim/telescope.py
+++ b/pyuvsim/telescope.py
@@ -110,7 +110,8 @@ class BeamList:
             self.check_consistency(force=force_check)
 
     def check_consistency(self, force: bool = False):
-        """Check the consistency of all beams in the list.
+        """
+        Check the consistency of all beams in the list.
 
         This checks basic parameters of the objects for consistency, eg. the ``beam_type``.
         It is meant to be manually called by the user.
@@ -126,6 +127,7 @@ class BeamList:
         ------
         BeamConsistencyError
             If any beam is inconsistent with the rest of the beams.
+
         """
         if self.string_mode:
             if not force:
@@ -173,12 +175,14 @@ class BeamList:
 
     @property
     def x_orientation(self):
-        """The x_orientation of all beams in list (if consistent).
+        """
+        Return the x_orientation of all beams in list (if consistent).
 
         Raises
         ------
         BeamConsistencyError
             If the beams in the list are not consistent with each other.
+
         """
         if not self.is_consistent:
             self.check_consistency(force=True)
@@ -215,12 +219,14 @@ class BeamList:
 
     @property
     def beam_type(self):
-        """The beam_type of all beams in list (if consistent).
+        """
+        Return the beam_type of all beams in list (if consistent).
 
         Raises
         ------
         BeamConsistencyError
             If the beams in the list are not consistent with each other.
+
         """
         if not self.is_consistent:
             self.check_consistency(force=True)
@@ -304,6 +310,7 @@ class BeamList:
 
         Converts objects to strings if in object mode,
         or vice versa if in string mode.
+
         """
         if self.string_mode:
             # If value is a string, then _scrape_uvb_params returns an empty dictionary.
@@ -346,6 +353,7 @@ class BeamList:
 
         Converts objects to strings if in object mode,
         or vice versa if in string mode.
+
         """
         if self.string_mode:
             self._str_beam_list.append('')
@@ -442,6 +450,7 @@ class BeamList:
         ValueError:
             If any UVBeam objects lack a "beam_path" keyword in the "extra_keywords"
             attribute, specifying the path to the beamfits file that generates it.
+
         """
         if self._obj_beam_list != []:
             # Convert object beams to string definitions
@@ -481,6 +490,8 @@ class BeamList:
 
 
 class BeamConsistencyError(Exception):
+    """An Exception class to mark inconsistencies among beams in a BeamList."""
+
     pass
 
 
--- a/pyuvsim/telescope.py
+++ b/pyuvsim/telescope.py
@@ -13,6 +13,7 @@ from . import mpi
 
 
 def consistent(method):
+    """Decorator for checking consistency before accessing a parameter."""
     @functools.wraps(method)
     def inner(self):
         if not self.is_consistent:
@@ -55,6 +56,15 @@ class BeamList:
     uvb_params : dict (optional)
         Options to set uvb_params, overriding settings from passed-in UVBeam objects.
 
+    check
+        Whether to perform a consistency check on the beams (i.e. asserting that several
+        of their defining parameters are the same for all beams in the list).
+
+    force_check
+        The consistency check is only possible for object-beams (not string mode). If
+        `force_check` is True, it will convert beams to object-mode to force the check
+        to run (then convert back to string mode).
+
     Raises
     ------
     ValueError :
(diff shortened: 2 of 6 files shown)


aelanman (commented):
Hi @steven-murray. Thanks for putting this together. It looks good to me. Once those missing lines are covered it should be ready to merge.

bhazelton:
This looks good, but I actually have some work I started on another branch that addresses the issue of the `x_orientation` and I'd like to figure out how that fits in with this. 

In my branch I added a check for the `x_orientation` of the beams in the BeamList and errored if they conflicted and warned if there was no `x_orientation` set for any of the beams that have that attribute. Then I used the `x_orientation` of the beams to set the `x_orientation` of the final UVData object. @aelanman I ran into some trouble with testing my code because it seems that the `x_orientation` is not preserved through the conversion to a string representation and back. I think this is because it's not getting captured in the `uvb_params` dict, but I can't tell where the definition is for what attributes of a UVBeam object are captured in that dict. I'd like to update it to capture x_orientation but I don't know how. Can you help me find the relevant bit of code there? Or do I need to update that in the file for it to work properly?

aelanman:
@bhazelton I think what you're looking for is the `_float_params` dictionary defined at the start of the `BeamList` class.

bhazelton:
@aelanman I don't think so, those are only used for analytic beams, I need something for UVBeams.

pyuvsim/telescope.py
```diff
@@ -128,30 +136,47 @@ def __init__(self, beam_list=None, uvb_params=None):
...
+                warnings.warn(
+                    "All polarized beams have x_orientation set to None. This will make it "
+                    "hard to interpret the polarizations of the simulated visibilities."
+                )
+
+    @consistent
+    @property
+    def x_orientation(self):
+        """The x_orientation of all beams in list (if consistent)."""
+        return getattr(self._obj_beam_list[0], "x_orientation", None)
```
bhazelton, line 153: What is returned if the beam list is not consistent? Does it error?

pyuvsim/tests/test_telescope.py
```diff
@@ -211,3 +211,37 @@ def test_beamlist_errors(beam_objs):
...
+def test_beamlist_consistency_stringmode(beam_objs):
+    newbeams = copy.deepcopy(beam_objs)
+    beamlist = pyuvsim.BeamList(newbeams[:2])
+    beamlist.set_str_mode()
+    beamlist.check_consistency(force=True)
+    assert beamlist.string_mode
+    with pytest.warns(UserWarning) as record:
+        beamlist.check_consistency(force=False)
+        if not record:
+            pytest.fail("Expected a warning!")
```
bhazelton, line 240: This is an interesting approach, why are you using it instead of the `match` keyword to `pytest.warns`?. Another option that I like even better is to use pyuvdata.test.check_warnings to check warnings -- it's got the same pattern of usage as `pytest.warns`, but it checks that the warnings that are specified exactly match (in number, type and message) the warnings raised by the code.

bhazelton (commented):
This is looking really good but I'd love a little bit of an explanation of some of the architecture here, I'm not sure I'm exactly following it. Maybe some illuminating comments in the code would help too.

pyuvsim/telescope.py
```diff
@@ -45,6 +46,13 @@ def __eq__(self, other):
             )
         return False
 
+def consistent(method):
+    @functools.wraps(method)
+    def inner(self):
+        if not self.is_consistent:
+            self.check_consistency(force=True)
+
+    return inner
```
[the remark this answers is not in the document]
steven-murray: Ah, I was trying to be clever, but I think I ended up not using this decorator anyway. I'll remove

pyuvsim/tests/test_telescope.py
steven-murray: oh nice, didn't know that!

steven-murray:
@bhazelton  thanks for the review: I'll incorporate those ideas. Unfortunately tests are failing locally atm, and I haven't had time to sort through it (and won't till Nov 15).

bhazelton:
ping on this @steven-murray

pyuvsim/telescope.py
steven-murray: Yes, it will error with `BeamConsistencyError` if not consistent. I can add that to the docstring.

pyuvsim/telescope.py
```diff
@@ -381,9 +411,77 @@ def set_obj_mode(self, use_shared_mem=False):
...
             self._obj_beam_list = [self._str_to_obj(bstr, use_shared_mem=use_shared_mem)
                                    for bstr in self._str_beam_list]
         self._set_params_on_uvbeams(self._obj_beam_list)
         self._str_beam_list = []
         self.string_mode = False
+        
+        if check:
+            self.check_consistency()
+
+    def check_consistency(self, check_pols: bool = True, force: bool=False):
```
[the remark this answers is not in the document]
steven-murray: Good idea. I've put it just after __init__

pyuvsim/telescope.py
```diff
@@ -381,9 +411,77 @@ def set_obj_mode(self, use_shared_mem=False):
...
+        """Check the consistency of all beams in the list.
+
+        This checks basic parameters of the objects for consistency, eg. the ``beam_type``.
+        It is meant to be manually called by the user.
+
+        Parameters
+        ----------
+        check_pols
```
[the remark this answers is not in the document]
steven-murray: Good catch.

steven-murray:
@bhazelton I've fixed up the comments you had. However, I'm getting quite a few test failures, due to beamlists in the tests that are not consistent (previously existing tests that were previously passing). I could turn off the checks in these tests, or I could update the beamlists, or change the code so that these are not considered "inconsistent". I don't think I'm familiar enough with the code to know which is best to do...

steven-murray:
If it helps, the majority of test failures are for `BeamConsistencyError: beam 2 does not have x_orientation but beam 1 does`

steven-murray:
@bhazelton this is ready for another review. The external herasim tests are failing because we explicitly create non-consistent beamlists to test there (we catch them in hera_sim, but once this PR is in, we can outsource those checks). 

There is one uncovered line in the diff, but I didn't add that line, so it must have been uncovered previously, but I just moved the line or something. Up to you if you want to get that covered in this PR.

pyuvsim/uvsim.py
```diff
@@ -427,6 +427,12 @@ def run_uvdata_uvsim(input_uv, beam_list, beam_dict=None, catalog=None, quiet=Fa
     # Construct beam objects from strings
     beam_list.set_obj_mode(use_shared_mem=True)
 
+    # In case the user created the beam list without checking consistency:
+    beam_list.check_consistency()
+
+    if beam_list.beam_type != 'efield':
+        raise ValueError("Beam type must be efield!")
```
bhazelton, line 434: I think we should cover this line. I don't see anything like it in the main branch, so I think it was added in this PR.

bhazelton (commented):
I think this is getting very close. Thanks for all the work on it @steven-murray !

steven-murray:
@bhazelton I think I have addressed all the comments, though it might need to be rebased on main.

bhazelton:
@steven-murray this looks good, but the tests are failing. Do you want me to do the rebase? I'm happy to.

steven-murray:
@bhazelton I rebased, and fixed the failing test that I botched. However, there are other failing tests that arise due to new warnings being emitted by distutils (nothing to do with this PR, but obviously an updated package?) Do you have ideas there? Feel free to go ahead and fix directly if you know what's happening.

bhazelton:
The PR I just merged fixes these distutils deprecations warnings that were breaking the tests. So I can rebase this if that's ok with you and I think the tests will be fixed.

steven-murray:
@bhazelton Go ahead! Thank you.

CHANGELOG.md
```diff
@@ -11,6 +11,7 @@
 ## [1.2.1] - 2021-10-13
 
 ### Added
+- New `check_consistency` method for `BeamList` objects.
```
bhazelton, line 14: This line needs to be moved up to under the Unreleased heading.

pyuvsim/telescope.py
```diff
@@ -45,6 +45,15 @@ class BeamList:
...
 
+    check
+        Whether to perform a consistency check on the beams (i.e. asserting that several
+        of their defining parameters are the same for all beams in the list).
+
+    force_check
+        The consistency check is only possible for object-beams (not string mode). If
+        `force_check` is True, it will convert beams to object-mode to force the check
+        to run (then convert back to string mode).
+
```
bhazelton, line 56: There should not be empty lines between these items in the Parameters section. We also usually indicate type here, but maybe that's not required because it's done in the function signature?

pyuvsim/telescope.py
```diff
@@ -89,31 +105,137 @@ def __init__(self, beam_list=None, uvb_params=None):
...
+            if not items:
+                # If none of the beams has this item, move on.
+                return
+
+            min_index = min(items.keys())
+
+            for indx, thing in items.items():
+                if np.any(thing != items[min_index]):
+                    raise BeamConsistencyError(
```
bhazelton, line 158: We should move the definition of the BeamConsistencyError class up to before this class, so it is defined before it is called.

bhazelton (commented):
Just a couple minor things to fix after the rebase.

pyuvsim/telescope.py
bhazelton: Looking at the RTD docs, I think we should add the types here because they don't get rendered the same way even though they are in the signature.

Screen Shot 2022-02-01 at 11 58 30 AM

pyuvsim/telescope.py
```diff
@@ -89,31 +105,137 @@ def __init__(self, beam_list=None, uvb_params=None):
...
+
+        This checks basic parameters of the objects for consistency, eg. the ``beam_type``.
+        It is meant to be manually called by the user.
+
+        Parameters
+        ----------
+        force
+            Whether to force the consistency check even if the object is in string
+            mode. This wil convert to to object mode, perform the check, then
+            convert back.
```
bhazelton, line 124: We should mark the type here as well.

pyuvsim/telescope.py
```diff
@@ -320,31 +450,50 @@ def set_str_mode(self):
...
 
         Sets :attr:`~.string_mode` to False.
+
+        Parameters
+        ----------
+        use_shared_mem
+            Whether to use shared memory for the beam data.
+        check
+            Whether to perform consistency checks on all the beams in the list after
+            conversion to object mode.
```
bhazelton, line 478: And here.

bhazelton (commented):
A couple more docstring suggestions.

bhazelton:
@steven-murray I just realized that you made `BeamList._obj_to_str` a class method but you left `BeamList._str_to_obj` as a regular method. Is that what you wanted to do? I don't really see why `BeamList._obj_to_str` needs to be a class method given the ways it's called, but it feels a weird to have this asymmetry between them.

steven-murray:
@bhazelton I couldn't get `_str_to_obj` to be a classmethod because it has a reference to `self` in it. I did feel icky about the asymmetry, and would feel better if both were class methods. I'm happy to take out the `@classmethod` from `_obj_to_str` if you prefer, or just leave the asymmetry.

bhazelton:
What is the added utility in making `_obj_to_str` a classmethod?

steven-murray:
Mostly it's a matter of principal -- I usually try to make things class methods if it's possible to. It means I don't _have_ to instantiate an object to use that functionality, which can come in handy later. For example, maybe one day we want to use this function to convert a beam object to/from obj/str where we've received that object from elsewhere. We could always create an instance of course, but it seems cleaner not to if we don't have to.

bhazelton:
Interesting. I tend to prefer to move things that do not need to be on an object (because e.g. they don't use the attributes) to functions rather than class methods. In this case, I think I'd prefer to leave them as regular methods. They are currently not part of the public API (they start with an underscore). If they become useful outside the BeamList object, I'd prefer to turn them into public functions instead.

steven-murray:
I can definitely understand that reasoning. OK, I have removed the classmethod decorator.

bhazelton (dismissed):
Looks good. Thanks!

bhazelton:
Darn, it can't be merged because of conflicts. Probably because of the merge commit. I'll fix that now.

bhazelton (approved):

PR #363, opened 2021-09-27
Add handling and checking for x_orientation

bhazelton:
## Description

Add a check on the x_orientation of the beams. If it is set consistently on all the beams, use it to set the output uvdata x_orientation. If it is inconsistent across the beams raise an error.

I also moved a function so that it was defined before it was called and I had to do a little refactoring to keep the complexity of `create_mock_catalog` under the limit.

## Types of changes

- [ ] Version change

## Checklist:

For all pull requests:

Bug fix checklist:

### DIFF
--- a/pyuvsim/telescope.py
+++ b/pyuvsim/telescope.py
@@ -128,6 +128,31 @@ class BeamList(object):
         self.uvb_params.update(list_uvb_params)
         self.uvb_params.update(uvb_params)    # Optional parameters for the UVBeam objects
 
+        x_orientation_list = []
+        for beam in beam_list:
+            if self.string_mode:
+                beam_use = self._str_to_obj(beam)
+            else:
+                beam_use = beam
+
+            if hasattr(beam_use, "x_orientation"):
+                x_orientation_list.append(beam_use.x_orientation)
+        if (
+            len(x_orientation_list) == 0
+            or x_orientation_list.count(None) == len(x_orientation_list)
+        ):
+            self.x_orientation = None
+        elif x_orientation_list.count(x_orientation_list[0]) == len(x_orientation_list):
+            self.x_orientation = x_orientation_list[0]
+        else:
+            raise ValueError("UVBeam x_orientations do not match among beams in list.")
+
+        if len(x_orientation_list) > 0 and self.x_orientation is None:
+            warnings.warn(
+                "All polarized beams have x_orientation set to None. This will make it "
+                "hard to interpret the polarizations of the simulated visibilities."
+            )
+
     def _scrape_uvb_params(self, beam_objs, strict=True):
         """
         Collect uvb_params from the set of beam objects.
--- a/pyuvsim/telescope.py
+++ b/pyuvsim/telescope.py
@@ -142,7 +142,7 @@ class BeamList(object):
             or x_orientation_list.count(None) == len(x_orientation_list)
         ):
             self.x_orientation = None
-        elif x_orientation_list.count(x_orientation_list[0]) == len(x_orientation_list):
+        elif all(x == x_orientation_list[0] for x in x_orientation_list):
             self.x_orientation = x_orientation_list[0]
         else:
             raise ValueError("UVBeam x_orientations do not match among beams in list.")
--- a/pyuvsim/tests/test_telescope.py
+++ b/pyuvsim/tests/test_telescope.py
@@ -3,14 +3,16 @@ import os
 import copy
 from astropy.coordinates import EarthLocation
 
-from pyuvdata import UVBeam
 import pytest
+from pyuvdata import UVBeam
+import pyuvdata.tests as uvtest
 
 import pyuvsim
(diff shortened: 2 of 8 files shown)


pyuvsim/telescope.py
```diff
@@ -128,6 +128,31 @@ def __init__(self, beam_list=None, uvb_params=None):
...
+            if self.string_mode:
+                beam_use = self._str_to_obj(beam)
+            else:
+                beam_use = beam
+
+            if hasattr(beam_use, "x_orientation"):
+                x_orientation_list.append(beam_use.x_orientation)
+        if (
+            len(x_orientation_list) == 0
+            or x_orientation_list.count(None) == len(x_orientation_list)
```
steven-murray, line 142: is it possible to have `None` for `x_orientation`? It seems to me that a null value should be _either_ be encoded as it not existing _or_ being `None`, but not both?

pyuvsim/telescope.py
```diff
@@ -128,6 +128,31 @@ def __init__(self, beam_list=None, uvb_params=None):
...
+        ):
+            self.x_orientation = None
+        elif x_orientation_list.count(x_orientation_list[0]) == len(x_orientation_list):
```
steven-murray, line 145: i get this syntax and it's fine, but I _prefer_ `if any(x != x_orientation_list[0] for x in x_orientation_list)` for semantic meaning.

pyuvsim/tests/test_telescope.py
```diff
@@ -181,6 +181,23 @@ def test_beamlist_errors(beam_objs):
     with pytest.raises(ValueError, match='Invalid file path'):
         beamlist.append(beam_path)
 
+    # test error on beams with different x_orientation
+    newbeams[0].x_orientation = None
+    with pytest.raises(
+        ValueError, match="UVBeam x_orientations do not match among beams in list."
```
steven-murray, line 187: I'm happy not to, but it might be worth considering using a custom error, like `BeamListConsistencyError`. It may be useful later to be able to catch such errors in certain contexts.

steven-murray (dismissed):
Thanks @bhazelton. I've provisionally approved, but made some suggestions (all of which are optional). 

The only other thing to think about would be how this interacts with #360, and whether to merge one before the other. I *think* we can get away with merging this first, then moving some of this logic to the method I've defined in #360.

pyuvsim/telescope.py
bhazelton: UVBeams (and possibly polarized analytic beams in the future) always have the x_orientation attribute and it is either set to `None` or to `east` or `north`. Unpolarized analytic beams do not have the x_orientation attribute and they don't need it because the two polarizations are identical. If a user is only using unpolarized analytic beams I didn't think they should get a warning.

steven-murray: hmmm, not sure I am fully on board with having different API's for the UVBeam/AnalyticBeam. I'd prefer the API to be the same and just to have `None` signify that the parameter doesn't exist. But since that is not the case, I'm happy with this code.

bhazelton: We can discuss this in future, we're planning to move AnalyticBeam to pyuvdata to better support it. I worry that there are two different meanings of None in that case -- One is that it's not defined but the two pols are different, the other is that there is no difference between the pols so it doesn't matter. Of course any realistic beam will have differences in the pol responses but I'm not sure we want to warn if the users are using beams that have no intrinsic polarization.

pyuvsim/telescope.py
bhazelton: but here I'm testing that they all are equal, so I think I want
`elif all(x == x_orientation_list[0] for x in x_orientation_list)`

pyuvsim/tests/test_telescope.py
bhazelton: Ok. I don't have strong feelings on this either way. I think we should talk about how this PR and PR #360 interact.

pyuvsim/telescope.py
steven-murray: sure thing.

pyuvsim/telescope.py
steven-murray: I see what you're saying. You might be right -- let's table the discussion for when `AnalyticBeam` moves to `pyuvdata`.

bhazelton:
@steven-murray the hera_sim tests appear to be failing on this PR but I'm not sure it's caused by the changes in this PR. Can you investigate?

aelanman (dismissed):
This looks fine to me. Thanks for putting this together!

bhazelton:
@steven-murray this PR is failing on the hera_sim tests because of a UVData select on polarization that is asking for the `"ee"` polarization but it looks like the UVData object doesn't have `x_orientation` set. I don't really see why this should have been affected by this PR, can you take a look?

steven-murray:
Ah. I'll look into it more, but I think this is something that will be dealt with properly in the upcoming simulator overhaul for hera_sim. At the moment, polarization for VisCPU is just broken on the main branch.

PR #364, opened 2021-09-27
Update Reference Sim Doc

### LINKED ISSUE #362: How do users get the Second Reference Sim data?
Been attempting to run the second generation reference sims for some tests. The Readme specifically mentions: 
- gleam.vot : mentions it is on google drive. Isn't this also what is downloaded with the `get_gleam.py` script? Should we mention that
- healpix_flat_EoR_spectrum_noise.hdf5 : Says it is available on the google drive but is not included in the google_drive_ids.dat file
-healpix_gsm_shell.hdf5 : same as above
- HERA_NicCST_fullfreq.uvbeam : Used in both the 1st and 2nd gens, mentions it is a UVBeam object in the docs but not where it is. No file with this name exists in either pyuvdata or pyuvsim data directories.

Like I mentioned before, none of this data seems to be downloaded by any of the scripts provided with the reference simulations.

mkolopanis:
closes #362 
## Description

Adds an optional argument to `download_latest_data.py` to get the uvbeam and two healpix skymodels from the google drive.
also adds a direct link to the google drive for users.

## Checklist:

For all pull requests:


bhazelton:
@steven-murray The hera_sim errors are happening on this PR too, which just updates documentation.

mkolopanis:
The error looks like there is no mpi installed, is hera_sim trying to call part of pyuvsim that needs mpi? maybe we need to update our herasim environment?

bhazelton:
oh, maybe! I think there were other errors besides that on the other PR, but I didn't check this one closely.

steven-murray:
Yup, looks like it's the MPI requirement. I wonder if it's possible to use hera_sim's own testing env instead of maintaining your own?

bhazelton (approved):
Looks good, thanks!

bhazelton (approved):
Looks good!

PR #371, opened 2021-10-25
Fix uvdata ordering

### LINKED ISSUE #370: Ordering of UVData matters
There seems to be an issue when the ordering of the input UVData is not "time". I came across this in testing for the hera_sim wrapper. The TLDR is that if you reorder the UVData object to anything but "time", then put it into pyuvsim, then reorder back to time, the resulting data array doesn't look right (eg. looking at data for non-redundant baselines gives the same visibilities, which can't be true). 

I'll put a not-quite-minimal failing example below, copied from the hera_sim tests. Note that we run this test over all our wrappers (eg. VisCPU and Healvis) and it passes for them.

```python

@pytest.fixture(scope="function")
def uvdata_linear():
    defaults.set("h1c")
    return io.empty_uvdata(
        Nfreqs=1,
        integration_time=sday.to("s") / NTIMES,
        Ntimes=NTIMES,
        array_layout={0: (0, 0, 0), 1: (10, 0, 0), 2: (20, 0, 0), 3: (0, 10, 0)},
        start_time=2456658.5,
        conjugation="ant1<ant2",
        polarization_array=["xx", "yy", "xy", "yx"],
    )

def test_ordering(uvdata_linear, simulator, order, conj):

    uvdata_linear.reorder_blts(order=order, conj_convention=conj)

    sky_model = make_point_sky(
        uvdata_linear,
        ra=np.linspace(0, 2 * np.pi, 8) * rad,
        dec=uvdata_linear.telescope_location_lat_lon_alt[0] * np.ones(8) * rad,
        align=False,
    )
       
    out_uv = pyuvsim.uvsim.run_uvdata_uvsim(
            input_uv=uvdata_linear,
            beam_list=<beamlist>,
            beam_dict=beam_dict,
            catalog=pyuvsim.simsetup.SkyModelData(sky_model),
     )
    uvdata_linear.data_array = out_uv.data_array

    uvdata_linear.reorder_blts(order="time", conj_convention="ant1<ant2")

    # THIS FAILS
    assert not np.allclose(
        uvdata_linear.get_data((0, 1)), uvdata_linear.get_data((0,3))
    )
```

mkolopanis:
Forces time-baseline ordering before a simulation.

## Motivation and Context

time-baseline ordering was assumed in the task index creation, this way we can enforce it.
Attempts to re-order back if the ordering was well defined before the simulation.
closes #370 

## Types of changes

- [ ] Version change

## Checklist:

For all pull requests:

Bug fix checklist:

### DIFF
--- a/benchmarking/benchmark.py
+++ b/benchmarking/benchmark.py
@@ -1,8 +1,9 @@
 import os
 import re
-import h5py
-import yaml
 import socket
+
+import yaml
+import h5py
 import numpy as np
 from pyradiosky import write_healpix_hdf5
 
--- a/benchmarking/benchmark.py
+++ b/benchmarking/benchmark.py
@@ -1,8 +1,9 @@
-import numpy as np
-import yaml
 import os
+import re
 import h5py
+import yaml
 import socket
+import numpy as np
 from pyradiosky import write_healpix_hdf5
 
 from pyuvsim.simsetup import _write_layout_csv, freq_array_to_params
@@ -233,12 +234,25 @@ def update_runlog(settings_dict, logfile='BENCHMARKS.log'):
 
     data_file = os.path.join(settings_dict['data_out'], 'benchmark.uvh5')
 
-    pyuvsim_version = ''
+    # Look for "pyuvsim verison: ??" but disregard any number of spaces between
+    # "pyuvsim" "version" the ":" and the version itself.
+    # \d+\.\d+\.\d+ ==> version ?.?.?
+    # \.dev(?:\d)*\+g\w{7} ==> optional subversion string too ".dev##+g" followed by the hash
+    # (\.\w+)? possibly a branch name too
+    # (?:.....)? means an match this group 0 or one times (essentially making it optional)
+    pattern = r"pyuvsim\s*version\s*:\s*(?P<VERSION>\d+\.\d+\.\d+(\.dev(?:\d)*\+g\w{7}(?:\.\w+)?)?)"
+
     with h5py.File(data_file, 'r') as dfile:
         hist = dfile['Header/history'][()].decode('UTF-8')
-        for line in hist.split('  '):    # Two spaces
-            if 'pyuvsim version' in line:
-                pyuvsim_version = line.split(': ')[-1]
+
+        match = re.search(pattern, hist)
+
+        if match is not None:
+            pyuvsim_version = match.group("VERSION")
+        else:
+            # try to just split it out of the history brute force.
+            # diregard the last character (a '.' in a sentence).
+            pyuvsim_version = hist.split("pyuvsim version: ")[1].split(" ")[0][:-1]
 
     header_vals = [
         "Date/Time", 'uvsim_version', 'HostName', 'SettingsFile',
--- a/pyuvsim/tests/test_uvsim.py
+++ b/pyuvsim/tests/test_uvsim.py
@@ -611,7 +611,7 @@ def test_task_coverage():
         tasks_all = []
         for rank in range(Npus):
             task_inds, src_inds, Ntasks_local, Nsrcs_local = pyuvsim.uvsim._make_task_inds(
-                Nbls, Ntimes, Nfreqs, Nsrcs, rank, Npus
+                Nbls * Ntimes, Nfreqs, Nsrcs, rank, Npus
             )
             src_inds = np.arange(Nsrcs)[src_inds]   # Turn slice into iterator
             tasks = itertools.product(task_inds, src_inds)
@@ -638,7 +638,7 @@ def test_task_coverage():
         tasks_all = []
         for rank in range(Npus):
             task_inds, src_inds, Ntasks_local, Nsrcs_local = pyuvsim.uvsim._make_task_inds(
-                Nbls, Ntimes, Nfreqs, Nsrcs, rank, Npus
+                Nbls * Ntimes, Nfreqs, Nsrcs, rank, Npus
             )
 
             tasks = itertools.product(task_inds, src_inds)
--- a/pyuvsim/tests/test_uvsim.py
+++ b/pyuvsim/tests/test_uvsim.py
@@ -80,7 +80,7 @@ def uvobj_beams_srcs():
     beam_list[0] = pyuvsim.AnalyticBeam('airy', diameter=13.0)
     beam_list.append(pyuvsim.AnalyticBeam('airy', diameter=14.6))
     beam_list.append(pyuvsim.AnalyticBeam('gaussian', sigma=0.3))
-    beam_list.append(pyuvsim.AnalyticBeam('gaussian', sigma=0.3,
+    beam_list.append(pyuvsim.AnalyticBeam('gaussian', sigma=0.8,
                      ref_freq=ref_freq, spectral_index=alpha))
 
     # Assign the last few antennas to use these other beams.
--- a/pyuvsim/tests/test_uvsim.py
+++ b/pyuvsim/tests/test_uvsim.py
@@ -925,18 +925,19 @@ def test_ordering(uvdata_two_redundant_bls_triangle_sources, order):
     uvdata_linear.reorder_blts(*order)
 
     out_uv = pyuvsim.uvsim.run_uvdata_uvsim(
-        input_uv=uvdata_linear,
+        input_uv=uvdata_linear.copy(),
         beam_list=beam_list,
         beam_dict=beam_dict,
         catalog=sky_model,
     )
-
     assert out_uv.blt_order == order
+    assert out_uv.blt_order == uvdata_linear.blt_order
 
     uvdata_linear.data_array = out_uv.data_array
 
-    uvdata_linear.reorder_blts(order="time", conj_convention="ant1<ant2")
-
-    assert not np.allclose(
+    assert np.allclose(
         uvdata_linear.get_data((0, 1)), uvdata_linear.get_data((1, 2))
+    )
+    assert not np.allclose(
+        uvdata_linear.get_data((0, 1)), uvdata_linear.get_data((0, 2))
     )
--- a/pyuvsim/tests/test_uvsim.py
+++ b/pyuvsim/tests/test_uvsim.py
@@ -946,6 +946,7 @@ def test_ordering(uvdata_two_redundant_bls_triangle_sources, order):
 
 
 def test_nblts_not_square(uvdata_two_redundant_bls_triangle_sources):
+    pytest.importorskip('mpi4py')
     uvdata_linear, beam_list, beam_dict, sky_model = uvdata_two_redundant_bls_triangle_sources
 
     uvdata_linear.conjugate_bls("ant1<ant2")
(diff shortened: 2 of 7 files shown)


pyuvsim/uvsim.py
```diff
@@ -280,12 +280,11 @@ def uvdata_to_task_iter(task_ids, input_uv, catalog, beam_list, beam_dict, Nsky_
...
 
     baselines = {}
-    Ntimes = input_uv.Ntimes
     Nfreqs = input_uv.Nfreqs
-    Nbls = input_uv.Nbls
+    Nblts = input_uv.Nblts
 
-    tasks_shape = (Ntimes, Nfreqs, Nbls)
-    time_ax, freq_ax, bl_ax = range(3)
+    tasks_shape = (Nblts, Nfreqs)
```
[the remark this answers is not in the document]
mkolopanis: I think we can get that same kind of behavior back, but it would require a re-evaluation of the make_task_inds and things like that. Basically it would need to iterate over unique times, then for each time find baseline which have that time, then iterate through each frequency. A little more logic mostly, just can't use the easy `unravel_index` from numpy.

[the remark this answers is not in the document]
aelanman: The difficulty here seems to be just in designing for the "non-square" case, when some times are missing baselines. I'm not sure if it would be faster to just drop computed visibilities that aren't requested, or to find some way to skip computing them.

I'm glad to see that you still require the `blt` ordering to be times then baselines. At no point should the final order of data in the file dictate the order of calculations within the task loop.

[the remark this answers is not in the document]
mkolopanis: I've been thinking about this, we always know that nblts * nfreqs arrays will be "square". what if we did something like this when task ids. It would require a memory overhead to generate the order array :thinking:  but it would force the ids to be iterated over in the time, baseline, freq order.
```
        _times = np.tile(input_uv.time_array, (input_uv.Nfreqs, 1)).T.flatten()[task_ids]
        _bls = np.tile(input_uv.baseline_array, (input_uv.Nfreqs, 1)).T.flatten()[task_ids]
        _freqs = np.tile(input_uv.freq_array.flatten(), (input_uv.Nblts, 1)).flatten()[task_ids]
        order = np.lexsort((_freqs, _bls, _times))

        # for task_index in task_ids:
        for index in order:
            task_index = task_ids[index]
```

mkolopanis:
Maybe? You would get Nbtlf float64s for time and freq, and Nbltf int64 for
baseline and order. So 256 x Nbltf bits of information. I think that is
1GiB of ram per 32768 tasks. Right would have to switch them in the sort.
But the idea is the same.

aelanman:
If it's happening in the `uvdata_to_task_iter` function, those arrays will be generated on each MPI process.

mkolopanis:
Yeah I understand this.

aelanman:
So wouldn't that be 1 GiB * Npus_per_node of memory used on each node? Once it downselects, it'll be less, but up front it will make an array of the size Nblts * Nfreqs on each process.

mkolopanis:
Yes, in the current prototype implementation that will happen. There may be
more sly ways to do it too. I was interested to see how the idea would be
received but it seems like we've begun focusing on details instead. I'll
post something else once I've found a well optimized solution.

mkolopanis:
Looks like I was off by an order of magnitude. it is 1MiB per 30k tasks. had a factor of 10 problem.

aelanman:
Your `Ntasks` is for 32 baselines?

mkolopanis:
I'm not sure I understand the question.
I had divided 2**20 by my estimation of the size per task instead of 2**30.

aelanman:
I don't understand where "32768" tasks came from earlier. If that's Nblts * Nfreqs, then for 1024 freqs that's Nblts = 32.

mkolopanis:
This is what I did. 
tasks to make 1 GiB of ram = 1 GiB / ( number of arrays * element size of array) 

or what I plugged into a python terminal for 4 arrays with 64 bit elements.
```python
tasks_for_1GIB = 2 ** 30 / ((4 * 64) / 8)
```

mkolopanis:
I've pushed up a new version with views, broadcasting, iterators, and slices where possible. I've empirically seen a 1e-5 to 3e-5 MiB increase ram usage per task per PU.

aelanman:
Excellent! This looks great. Thanks for figuring that out!

bhazelton:
@steven-murray Do you understand the herasim test failures on this PR? Also, I think it's getting close to completion if you want to have another look.

steven-murray:
@bhazelton I don't think the breaking tests have anything to do with pyuvsim per se, and they will likely be fixed soon with a major PR in hera_sim. I think you can merge without those tests passing.

pyuvsim/tests/test_uvsim.py
```diff
@@ -101,6 +101,26 @@ def uvobj_beams_srcs():
     return uv_obj, beam_list, beam_dict, sources
 
 
+@pytest.fixture
+def uvdata_two_redundant_bls_triangle_sources():
+    # A uvdata object, beam list, beam dict, and source array.
```
steven-murray, line 106: ... walked into a bar.

pyuvsim/uvsim.py
```diff
@@ -532,6 +572,15 @@ def run_uvdata_uvsim(input_uv, beam_list, beam_dict=None, catalog=None, quiet=Fa
...
     if rank == 0:
+        if input_order is not None:
+            if len(input_order) < 2:
+                input_order = (input_order[0], None)
+
+            uv_container.reorder_blts(
+                order=input_order[0],
+                minor_order=input_order[1],
+            )
+
```
steven-murray, line 583: So, here you reorder back to the original order if it was set. I can see two potential problems (which may not actually be problems:

1. If `input_order` is None, you don't re-order back to original. So the output order will be different from input. Like, if user puts in something with ridiculous arbitrary order, you just go ahead and reorder it and give it back in (time, baseline). Not sure there's any way around this, but it should at least be well-documented (or warned).
2. You reorder even if the input order was already (time, baseline). This is a waste of time, unless the reorder_blts function itself short-circuits it.

steven-murray (dismissed):
Only one minor comment which you can take or leave. Thanks all!

pyuvsim/uvsim.py
mkolopanis: We could issue a warning if the input order is not known and tell the  output will be (time, baseline). Or just scream it in the docstrings, or both.

good catch on 2, it looks like it does not short-circuit so that could be a decent waste of time.

pyuvsim/uvsim.py
```diff
@@ -532,6 +572,22 @@ def run_uvdata_uvsim(input_uv, beam_list, beam_dict=None, catalog=None, quiet=Fa
...
+            if input_order != ("time", "baseline"):
+                uv_container.reorder_blts(
+                    order=input_order[0],
+                    minor_order=input_order[1],
+                )
+        else:
+            warnings.warn(
+                "The parameter `blt_order` could not be identified for input_uv. "
+                "The ordering was changed to `(time, baseline)` during the simulation "
+                " but cannot be changed back."
```
bhazelton, line 588: I don't like the phrase "but cannot be changed back" because it might suggest to users that blt ordering can't be changed, which of course it can be. Maybe something like this instead:
"The parameter `blt_order` could not be identified for input_uv so the ordering cannot be restored. The output will be in `(time, baseline)` order."

mkolopanis: I've updated the wording to be more in line with your suggestion.

mkolopanis:
Here's an update on the benchmark from `main` and from the current version of this branch. Should I just add it to the base file even though it is untracked?
|Date/Time | uvsim_version  | HostName  | SettingsFile  | cpus-per-task  | Ntasks  | Nnodes  |MemLimit  |  Ntimes | Nbls | Nfreqs | Nsrcs  | Nsrcs_part |  MaxRSS [GiB] |   Runtime|
| :-- | :---  | :---  | :---  | :----:  |  :----:  |  :----:  | :----:  |   :----: |  :----: |  :----: |  :----:  |  :----: |  ---: |    :----:|
|2020-03-06-21:40:47.584612 | 1.1.2.       |    node474.oscar.ccv.brown.edu  | settings.yaml    |2|             10      |1|         50G         |21|        500     |256|       3072    |1|             13.071    |          4:32:16.373965|
|2020-03-26-17:54:43.959585 | 1.1.2.       |    node474.oscar.ccv.brown.edu  | settings.yaml    |2|             10      |1|         50G         |21|        500     |256|       3072    |1|             12.187   |           3:24:00.635442|
|2020-04-02-15:50:16.314578 | 1.1.2.       |    node473.oscar.ccv.brown.edu  | settings.yaml    |2|             10      |1|         50G         |21|        500     |256|       3072    |1|             11.475  |            0:10:30.647003|
|2021-11-08-13:29:18.779374 | 1.2.2.dev1+g23f68a4.main  |herapost-master.aoc.nrao.edu   |settings.yaml  |2              |10     |1      |50G        |21     |500    |256    |3072   |1          |2.053              |0:11:56.878661|
|2021-11-08-11:38:08.304418 | 1.2.2.dev25+gef5c800.fix_uvdata_ordering  |herapost-master.aoc.nrao.edu   |settings.yaml  |2              |10     |1      |50G        |21     |500    |256    |3072   |1              |2.061     |    0:11:46.983807|

mkolopanis:
I'm also running reference sim 2.1 on the NRAO cluster to check timings. I'll post an update when I have one.

mkolopanis:
with an N of 1, reference sim 2.1 execution time on the NRAO cluster differed by 6 minutes (longer on this branch). I don't think that's really an appreciable difference.

steven-murray:
Really, that's better. Gives you time to make coffee.

benchmarking/benchmark.py
```diff
@@ -1,8 +1,9 @@
-import numpy as np
-import yaml
 import os
+import re
 import h5py
+import yaml
 import socket
+import numpy as np
```
bhazelton, line 6: Let's split these, first listing standard lib packages then an empty line and then other packages.

benchmarking/benchmark.py
```diff
@@ -233,12 +234,25 @@ def update_runlog(settings_dict, logfile='BENCHMARKS.log'):
...
     data_file = os.path.join(settings_dict['data_out'], 'benchmark.uvh5')
 
-    pyuvsim_version = ''
+    # Look for "pyuvsim verison: ??" but disregard any number of spaces between
+    # "pyuvsim" "version" the ":" and the version itself.
+    # \d+\.\d+\.\d+ ==> version ?.?.?
+    # \.dev(?:\d)*\+g\w{7} ==> optional subversion string too ".dev##+g" followed by the hash
+    # (\.\w+)? possibly a branch name too
+    # (?:.....)? means an match this group 0 or one times (essentially making it optional)
+    pattern = r"pyuvsim\s*version\s*:\s*(?P<VERSION>\d+\.\d+\.\d+(\.dev(?:\d)*\+g\w{7}(?:\.\w+)?)?)"
```
bhazelton, line 243: Wow. Thanks for the comments!!

bhazelton (commented):
Looks great, other than a minor comment about import ordering.

benchmarking/benchmark.py
mkolopanis: Think I moved them to the desired order now.

bhazelton (approved):
Looks good, thank you for all the hard work!

PR #373, opened 2021-11-02
Make --with-inputs a proper flag not a boolean input.

mkolopanis:
the script to download the reference data and inputs should have use a `store_true` action for the`--with-inputs` command.


bhazelton (approved):
Very straightforward fix. Thanks!

bhazelton:
I think this can be merged even with the herasim tests failing. Another test failed with a weird error, I just restarted that workflow, if all those tests pass it can be merged.

PR #390, opened 2022-03-14
Add return options to `initialize_uvdata_from_params` and `initialize_catalog_from_params`

### LINKED ISSUE #215: setting uv_obj as a tuple of stuff
Here: 

https://github.com/RadioAstronomySoftwareGroup/pyuvsim/blob/b10d0af1dea19af672a34f548d2be3e566d74da2/pyuvsim/simsetup.py#L1163

the return of the function has a `UVData` object, but also a beam list and beam ids. So the result of the function is a tuple, rather than the stated UVData object itself.

bhazelton:
## Description

- Add a `return_beams` parameter to the `initialize_uvdata_from_params` function to have it return the beam list and dict. Defaults to True currently, but will default to False starting in version 1.4.
- Add a `return_catname` parameter to the `initialize_catalog_from_params ` function to have it return the catalog name. Defaults to True currently, but will default to False starting in version 1.4.
- Also broke out some code into a helper function because of complexity issues.

## Motivation and Context

closes #215 

## Types of changes

- [ ] Version change

## Checklist:

For all pull requests:

New feature checklist:


bhazelton:
@steven-murray This PR addresses an issue you made some time ago. Can you take a look?

pyuvsim/simsetup.py
```diff
@@ -836,7 +836,77 @@ def _sky_select_calc_rise_set(sky, source_params, telescope_lat_deg=None):
...
+    elif filename.endswith('vot'):
+        if 'gleam' in filename:
+            if "spectral_type" in source_params:
+                spectral_type = source_params["spectral_type"]
+            else:
+                warnings.warn("No spectral_type specified for GLEAM, using 'flat'.")
+                spectral_type = "flat"
+            sky.read_gleam_catalog(
+                filename, spectral_type=spectral_type
+            )
```
steven-murray, line 869: this seems somewhat dangerous... what if the file is GLEAM but named something like "from_natasha.vot"? I guess the user can always just read it themselves if they have a weird special case. Maybe we should at least use `'gleam' in filename.lower()`

pyuvsim/simsetup.py
```diff
@@ -836,7 +836,77 @@ def _sky_select_calc_rise_set(sky, source_params, telescope_lat_deg=None):
...
+                        f"No {param} specified for {filename}, using default: {default}."
+                    )
+                    vo_params[param] = default
+                else:
+                    vo_params[param] = source_params[param]
+
+            vo_params["reference_frequency"] = None
+            sky.read_votable_catalog(filename, **vo_params)
+    elif filename.endswith('skyh5'):
+        sky.read_skyh5(filename)
```
steven-murray, line 890: In the future, it'd be good to have a `is_skyh5` function in pyradiosky that could tell you whether it was skyh5 for any input file. Because really you could have any extension and it still be valid.

steven-murray (approved):
Thanks @bhazelton and sorry for the slow response. This looks good to me -- I made two comments but you can take or leave them.

pyuvsim/simsetup.py
bhazelton: I added an optional `filetype` parameter to this method (and the one that calls it) to allow users to specify the filetype rather than relying on this parsing. It falls back on this parsing if the filetype is not specified.

I also added the `.lower()` for checking for gleam.

steven-murray (approved):
Looks good to me. I agree @bhazelton, having the more general `from_file` in `SkyModel` would be a good way forward.

bhazelton:
@steven-murray I just updated the changelog and docs, can you re-review?

pyuvsim/simsetup.py
bhazelton: The issue already exists! https://github.com/RadioAstronomySoftwareGroup/pyradiosky/issues/121

steven-murray (approved):

PR #394, opened 2022-03-30
generalizing jones matrix for Nfeeds=1,2

slosar:
Changes the jones matrix routine to (Nfeeds,2) instead of (2,2) generalizing to any Nfeeds. Down the line Nfeeds can still be only 1 or 2, but this can be changed. This is only a proposal.

## Description
See file changes - pretty obvious. Still specialized for theta-phi coordinate system.

### DIFF
--- a/pyuvsim/antenna.py
+++ b/pyuvsim/antenna.py
@@ -43,7 +43,7 @@ class Antenna:
         """
         Calculate the jones matrix for this antenna in the direction of sources.
 
-        A 2 x 2 x Nsources array of Efield vectors in Az/Alt.
+        A Nfeeds x 2 x Nsources array of Efield vectors in Az/Alt.
 
         Parameters
         ----------
@@ -67,7 +67,7 @@ class Antenna:
         Returns
         -------
         jones_matrix : array_like of float
-            Jones matricies for each source location, shape (2,2, Ncomponents). The
+            Jones matricies for each source location, shape (Nfeeds,2, Ncomponents). The
             first axis is feed, the second axis is vector component on the sky in az/za.
 
         """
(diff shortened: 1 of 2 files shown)


bhazelton:
@slosar @nmahesh1412 Sorry for the slowness, but I think this is now in a place where it could be merged. You are the two users I know of who are most interested in this, so I wanted to see if either of you wanted to try it out and see if it works for you.

slosar:
Hi Bryna, huh, I have completely forgotten about this, it was two years ago! :) In the mean time we have developed our own (harmonic-space) code to do the sims we need to do, so I do not need this feature and not even sure what to test. Perhaps @nmahesh1412 can help. I still think it is probably a good feature to have.

src/pyuvsim/uvsim.py
```diff
@@ -786,11 +792,37 @@ def run_uvdata_uvsim(
     if not isinstance(input_uv, UVData):
         raise TypeError("input_uv must be UVData object")
 
+    if beam_list.beam_type != "efield":
```
jpober, line 795: The codecov report says the repo remains 100% covered, but the file-by-file view suggests there are a number of lines (14 in total) here in uvsim.py that aren't covered by tests.

bhazelton: Yeah, these annotations are not accurate and are really annoying. Codecov is actually deprecating them because there's a limitation in the GitHub API that prevents codecov from updating or deleting the annotations after more CI jobs have run and the GitHub API also won't let codecov tell when all the runs have finished, so codecov cannot just wait until they're all done. Details here: https://docs.codecov.com/docs/github-checks

I could turn them off by adding a `.codecov.yml` file or we can just wait until codecov kills them. You can also toggle them off when you're reading a PR by pressing `a`.

jpober (approved):
Ah, that's great, thanks!

PR #401, opened 2022-05-09
Fix autos

### LINKED ISSUE #399: Non-real values of auto-correlations failed to pass the check of pyuvdata
I have ran into an error when writing the outputs of pyuvsim to disk. 

`ValueError: Some auto-correlations have non-real values in data_array. You can attempt to fix this by setting fix_autos=True.`

It seems pyuvsim.utils.write_uvdata has no controls on the parameter fix_autos (fix_autos is defaultly set to be false otherwise).

bhazelton:
## Description

Use the new `fix_autos` keyword in pyuvdata to fix autos to be real on file write-out (if the pyuvdata version is high enough).

Also fix some test errors associated with recent pyuvdata updates and fill in some missing recent info from the changelog.

## Motivation and Context

fixes #399 

## Types of changes

- [ ] Version change

## Checklist:

For all pull requests:

New feature checklist:

### DIFF
--- a/pyuvsim/tests/test_run.py
+++ b/pyuvsim/tests/test_run.py
@@ -49,6 +49,15 @@ def test_run_paramfile_uvsim(goto_tempdir, paramfile):
     uv_ref.read_uvfits(os.path.join(SIM_DATA_PATH, 'testfile_singlesource.uvfits'))
     uv_ref.unphase_to_drift(use_ant_pos=True)
     uv_ref.reorder_blts("time", minor_order="baseline")
+    # This is an old file with the bug that added one to the
+    # antenna numbers for uvfits files. Fix them (if pyuvdata is recent)
+    if np.min(np.union1d(uv_ref.ant_1_array, uv_ref.ant_2_array)) > 0:
+        uv_ref.ant_1_array = uv_ref.ant_1_array - 1
+        uv_ref.ant_2_array = uv_ref.ant_2_array - 1
+        uv_ref.antenna_numbers = uv_ref.antenna_numbers - 1
+        uv_ref.baseline_array = uv_ref.antnums_to_baseline(
+            uv_ref.ant_1_array, uv_ref.ant_2_array
+        )
 
     # set the x_orientation
     uv_ref.x_orientation = "east"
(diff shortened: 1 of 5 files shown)


pyuvsim/tests/test_run.py
```diff
@@ -49,6 +49,15 @@ def test_run_paramfile_uvsim(goto_tempdir, paramfile):
     uv_ref.read_uvfits(os.path.join(SIM_DATA_PATH, 'testfile_singlesource.uvfits'))
     uv_ref.unphase_to_drift(use_ant_pos=True)
     uv_ref.reorder_blts("time", minor_order="baseline")
+    # This is an old file with the bug that added one to the
+    # antenna numbers for uvfits files. Fix them (if pyuvdata is recent)
+    if np.min(np.union1d(uv_ref.ant_1_array, uv_ref.ant_2_array)) > 0:
```
mkolopanis, line 54: Is the uvfits bug the _only_ way this can happen though? ant_1_array and ant_2_array should be indices correct? So the answer to my first question should be yes?

mkolopanis (commented):
Seems pretty good. I have one small and one large question. Does this mean that users who do not have pyuvdata>=2.2.7 and end up with a complex auto just cannot write out a file via UVData?

pyuvsim/tests/test_run.py
bhazelton: They're not indices, they're antenna numbers. But until recently we were subtracting 1 from them (as well as from the antenna_numbers) when we wrote out uvfits files. That and a bug in our baseline number calculation was fixed in: https://github.com/RadioAstronomySoftwareGroup/pyuvdata/pull/1148

mkolopanis: Yeah I remember that, just wondering if the check posted here will only trigger off that bug right? or are there other ways for this to get triggered even if we didn't want it?

bhazelton: I would not use this generically, it's tuned for the particular file used in this test.

bhazelton: Antenna numbers can generally be zero, but this file is made from an antenna layout csv that has antenna numbers starting at 1.

mkolopanis: Ah yeah great okay. Tests. I must be distracted.

mkolopanis (approved):

PR #404, opened 2022-05-25
Add support to turn off az/za domain checking

### LINKED ISSUE #388: Add handling of check_azza_domain parameter for UVBeams
To allow for faster interpolation. See https://github.com/RadioAstronomySoftwareGroup/pyuvdata/pull/1144

bhazelton:
## Description

This adds a parameter to `run_uvsim` and `run_uvdata_uvsim` to allow for the check that the beam covers the interpolation location to be turned off. The default behavior is to turn off the check if the beam covers the full sky horizon to horizon.

Turning off this check can lead to substantial speedups, but if sources are simulated outside the beam area the response will be incorrect.

This requires updating out pyuvdata requirement to >= 2.2.8 when the `check_azza_domain` parameter was added to `UVBeam.interp`. I don't see a downside to updating this requirement.

## Motivation and Context

closes #388 

## Types of changes

- [ ] Version change

## Checklist:

For all pull requests:

New feature checklist:

### DIFF
--- a/pyuvsim/analyticbeam.py
+++ b/pyuvsim/analyticbeam.py
@@ -120,9 +120,7 @@ class AnalyticBeam:
         az_array,
         za_array,
         freq_array,
-        reuse_spline=None,
-        spline_opts=None,
-        check_azza_domain=None
+        **kwargs,
     ):
         """
         Evaluate the primary beam at given sky coordinates and frequencies.
@@ -140,11 +138,7 @@ class AnalyticBeam:
             same length as `az_array`.
         freq_array : array-like of float
             Frequency values to evaluate at in Hz. Should be a 1D array.
-        reuse_spline : bool
-            Unused. Here for compatibility with :meth:`pyuvdata.UVBeam.interp`.
-        spline_opts : dict
-            Unused. Here for compatibility with :meth:`pyuvdata.UVBeam.interp`.
-        check_azza_domain : dict
+        kwargs : dict
             Unused. Here for compatibility with :meth:`pyuvdata.UVBeam.interp`.
 
         Returns
--- a/pyuvsim/analyticbeam.py
+++ b/pyuvsim/analyticbeam.py
@@ -115,7 +115,15 @@ class AnalyticBeam:
         pol_strings = ['XX', 'XY', 'YX', 'YY']
         self.polarization_array = np.array([uvutils.polstr2num(ps.upper()) for ps in pol_strings])
 
-    def interp(self, az_array, za_array, freq_array, reuse_spline=None, spline_opts=None):
+    def interp(
+        self,
+        az_array,
+        za_array,
+        freq_array,
+        reuse_spline=None,
+        spline_opts=None,
+        check_azza_domain=None
+    ):
         """
         Evaluate the primary beam at given sky coordinates and frequencies.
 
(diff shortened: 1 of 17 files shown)


pyuvsim/analyticbeam.py
```diff
@@ -115,7 +115,15 @@ def efield_to_power(self):
...
 
-    def interp(self, az_array, za_array, freq_array, reuse_spline=None, spline_opts=None):
+    def interp(
+        self,
+        az_array,
+        za_array,
+        freq_array,
+        reuse_spline=None,
+        spline_opts=None,
+        check_azza_domain=None
```
steven-murray, line 125: Should we not just use `**kwargs` in the signature and then just ignore them? I imagine there might be more additions to `interp()` in pyuvbeam over time, and we don't always want to have to remember to add the parameter to the analytic beam.

pyuvsim/utils.py
```diff
@@ -270,24 +267,19 @@ def write_uvdata(
     if noclobber:
         outfile_name = check_file_exists_and_increment(outfile_name)
 
-    if version.parse(pyuvdata.__version__) >= version.parse("2.2.7"):
-        write_kwargs = {'fix_autos': fix_autos}
-    else:
-        write_kwargs = {}
-
     print('Outfile path: ', outfile_name, flush=True)
```
steven-murray, line 270: I know this isn't a part of this PR, but I don't like bare print statements. Should this be a logging statement? At least then the user can control its output.

steven-murray (commented):
Thanks @bhazelton, this is well done. I have only a couple of small comments. Looks like coverage needs improvement as well.

pyuvsim/uvsim.py
```diff
@@ -572,11 +590,30 @@ def _check_ntasks_valid(Ntasks_tot):
         )
 
 
+def _check_all_azza_beams_full_sky(beam_list):
```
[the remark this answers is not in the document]
bhazelton: Sure, I didn't think about it being more widely useful. I was also debating moving it to the BeamList object.

pyuvsim/utils.py
bhazelton: probably, I'll take a look at it.

pyuvsim/analyticbeam.py
bhazelton: sure, that's probably reasonable. Longer term this should move to pyuvdata and we can think better about how to handle these there.

pyuvsim/uvsim.py
```diff
@@ -572,11 +590,30 @@ def _check_ntasks_valid(Ntasks_tot):
...
+    for beam in beam_list:
+        if isinstance(beam, UVBeam) and beam.pixel_coordinate_system == "az_za":
+            # regular gridding is enforced in UVBeam checking, so can use first diff
+            axis1_diff = np.diff(beam.axis1_array)[0]
+            axis2_diff = np.diff(beam.axis2_array)[0]
+            max_axis_diff = np.max([axis1_diff, axis2_diff])
+            if not (
+                np.max(beam.axis2_array) >= np.pi / 2. - max_axis_diff * 2.0
+                and np.min(beam.axis1_array) <= max_axis_diff * 2.0
+                and np.max(beam.axis1_array) >= 2. * np.pi - max_axis_diff * 2.0
```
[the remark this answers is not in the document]
bhazelton: Actually, the regular grid requirement in the UVBeam check may ensure that there are no holes.

bhazelton:
@steven-murray I think I addressed your comments, but now the hera_sim tests are failing and I don't understand why. The error message is
```
TypeError: interp() got an unexpected keyword argument 'check_azza_domain'
```
Which should not be happening because the pyuvdata version is 2.2.8, which includes the change.

steven-murray (approved):

Issue #395, opened 2022-04-05
Properly handle UVBeams with E-field components not aligned with azimuth & zenith angle


bhazelton:
The code that takes the UVBeam data array into the jones matrix (in antenna.py) assumes a beam that has E-field components aligned with the azimuth and zenith angle axes. This should not be assumed, instead we need to use the `basis_vector_array` on UVBeam objects to get the mapping to azimuth/zenith angle aligned vectors.

PR #405, opened 2022-05-25
Check that beam basis vectors are aligned with the azimuth and zenith angle

bhazelton:
## Description

This adds a check that the beam basis vectors are aligned with the azimuth and zenith angle in each pixel. This is currently assumed but not checked in pyuvsim. 

## Motivation and Context

In the future we'd like to properly support other basis vectors (#395), but this ensures that we at least test for this and error if it's not the case.

## Types of changes

- [ ] Version change

## Checklist:

For all pull requests:

Bug fix checklist:

### DIFF
--- a/pyuvsim/telescope.py
+++ b/pyuvsim/telescope.py
@@ -116,6 +116,36 @@ class BeamList:
         if check:
             self.check_consistency(force=force_check)
 
+    def _get_beam_basis_type(self):
+        if self.string_mode:
+            raise ValueError("Cannot get beam basis type a string-mode BeamList.")
+
+        beam_basis = {}
+        for index, bm in enumerate(self):
+            if isinstance(bm, UVBeam):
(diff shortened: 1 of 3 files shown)


pyuvsim/telescope.py
```diff
@@ -113,6 +116,36 @@ def __init__(
         if check:
             self.check_consistency(force=force_check)
 
+    def _get_beam_basis_type(self):
+        if self.string_mode:
+            raise ValueError("Cannot get beam basis type a string-mode BeamList.")
```
steven-murray, line 121: missing "from"

steven-murray (dismissed):
All looks good except one typo, thanks @bhazelton !

pyuvsim/telescope.py
bhazelton: Thanks for catching that!

steven-murray (approved):

PR #417, opened 2023-02-06
Compatibility with pyradiosky>=0.2

bhazelton:
## Description

Mostly this required changing over to the new `skycoord` parameter on `SkyModel` objects. I decided to just make a clean break and require pyradiosky>=0.2 because it made the code much cleaner. In order to make this work with pyradiosky v0.3, I had to remove some code that called functions/options in pyradiosky that were removed in v0.3. I think we should bump the version of pyuvsim to v1.3 when we merge this PR because of those changes.

Note: there is one change in pyradiosky that required updating some test files: pyradiosky now interprets `J2000` in text file and votable (e.g. Gleam) catalog column headings as implying that the coordinate frame is `FK5` rather than `ICRS` (as it used to). This is more correct but it could lead to small changes compared to previous pyuvsim runs. This needs some careful testing.

 I also had to improve the lunarsky importing structure to match pyuvdata and pyradiosky -- this is required because it caused errors related to the mismatch between pyuvsim and the newest pyradiosky.

This also increased the version of lunarsky to get rid of astropy deprecation warnings that were cropping up.

## Types of changes

- [ ] Version change

## Checklist:

For all pull requests:

Bug fix checklist:


bhazelton:
@steven-murray heads-up, the hera_sim tests are failing on this PR.

steven-murray:
@bhazelton I *think* Bobby has fixes for this in https://github.com/HERA-Team/hera_sim/pull/248

pyuvsim/tests/test_simsetup.py
```diff
@@ -375,13 +394,22 @@ def test_healpix_hdf5_catalog():
     """
     pytest.importorskip('astropy_healpix')
     path = os.path.join(SKY_DATA_PATH, 'healpix_disk.hdf5')
-    sky = pyradiosky.SkyModel()
+    sky = SkyModel()
```
steven-murray, line 397: is there a reason we're not using `.from_file()` here?

pyuvsim/uvsim.py
```diff
@@ -538,7 +539,14 @@ def uvdata_to_task_iter(task_ids, input_uv, catalog, beam_list, beam_dict, Nsky_
         raise ValueError("If world keyword is set, it must be either 'moon' or 'earth'.")
     telescope = Telescope(input_uv.telescope_name, location, beam_list)
     freq_array = input_uv.freq_array * units.Hz
-    time_array = Time(input_uv.time_array, scale='utc', format='jd', location=telescope.location)
+    if hasmoon and isinstance(location, MoonLocation):
```
steven-murray, line 542: this might be a little neater as:

```python
tclass = LTime if (hasmoon and isinstance(location, MoonLocation) else Time
time_array = tclass(....)
```

steven-murray (commented):
This all looks reasonable to me. I only made a few simple comments.

pyuvsim/data/single_source.txt
```diff
@@ -1,3 +1,3 @@
-SOURCE_ID   RA_J2000    Dec_J2000   Flux
+SOURCE_ID   RA_icrs    Dec_icrs   Flux
```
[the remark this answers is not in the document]
bhazelton: doesn't matter, the check code is case-insensitive.

pyuvsim/uvsim.py
bhazelton: Agreed, except that I will keep the if/else on separate lines because I find that much more readable.

pyuvsim/tests/test_simsetup.py
bhazelton: Yes, there is. This is using an old, non-skyh5 file. We currently still have a reader for that in SkyModel, but it is deprecated and therefore it's not supported via the generic read or the `from_file` method. I considered removing this test, but I left it in for now since the support for these files does still exist in pyradiosky. But I could be convinced to remove it.

steven-murray: Ah ok cool.

steven-murray (dismissed):
Thanks @bhazelton this all looks good to me. I suppose we can merge, but still need to fix the ref sims before making a new release.

steven-murray (approved):
This all looks good to me. It's pretty large, so it's very possible I missed something, but nothing stands out as being wrong.

PR #424, opened 2023-04-01
Fix Nsky parts bug

bhazelton:
## Description

This fixes a major bug that was introduced between v1.2.1 and v1.2.2. It affects runs where the source list is large enough that it is broken across more than one processing unit.

I also had to pin pyuvdata to less than v2.3 because of our current pinning of pyradiosky.

## Types of changes

- [ ] Version change

## Checklist:

For all pull requests:

Bug fix checklist:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,9 @@
 - Use future array shapes on UVBeam objects.
 - Require pyuvdata >= 2.2.10
 
+### Fixed
+- A major bug introduced between v1.2.1 and v1.2.2 that caused errors in simulations
+when the source list was large enough that it needed to be split among processing units.
 
 ### Added
 - `reorder_kw` option to `initialize_uvdata_from_params` function.
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,7 +6,10 @@
 - Initial ordering of blt axis in `initialize_uvdata_from_params`
   (unchanged output by default).
 - Use future array shapes on UVBeam objects.
-- Require pyuvdata >= 2.2.10
+- Update dependency requirements to: pyuvdata >= 2.2.10,<2.3, pyradiosky>=0.1.0,<0.2,
+numpy>=1.19, scipy>=1.3
+- Update optional dependency requirements to: astropy-healpix>=0.6, lunarsky>=0.1.2,
+python-casacore>=3.3.1
 
 ### Fixed
 - A major bug introduced between v1.2.1 and v1.2.2 that caused errors in simulations
--- a/pyuvsim/__init__.py
+++ b/pyuvsim/__init__.py
@@ -2,14 +2,14 @@
 # Copyright (c) 2018 Radio Astronomy Software Group
 # Licensed under the 3-clause BSD License
 """Define namespace."""
+from importlib.metadata import PackageNotFoundError, version
 from pathlib import Path
 
-from pkg_resources import DistributionNotFound, get_distribution
 from setuptools_scm import get_version
 
 from .branch_scheme import branch_scheme
 
-try:  # pragma: nocover
+try:
     # get accurate version for developer installs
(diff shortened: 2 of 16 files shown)


CHANGELOG.md
```diff
@@ -8,6 +8,9 @@
 - Use future array shapes on UVBeam objects.
 - Require pyuvdata >= 2.2.10
 
+### Fixed
+- A major bug introduced between v1.2.1 and v1.2.2 that caused errors in simulations
+when the source list was large enough that it needed to be split among processing units.
```
bhazelton, line 14: Add dependency changes here.

bhazelton: done

pyuvsim/uvsim.py
```diff
@@ -98,6 +98,14 @@ def __init__(self, sources, time, freq, baseline, telescope, freq_i=0):
         if sources.spectral_type == 'flat':
             self.freq_i = 0
 
+    def __repr__(self) -> str:
+        """Make a nice representation."""
+        return (
+            f"UVTask, [time: {self.time}, freq: {self.freq}, source names: "
```
mkolopanis, line 104: My only thought here is do we want `UVTask[ .....]` without the ',' or with '<' or '('

mkolopanis (commented):
Looking good from my end. I have a style question about the repr

pyuvsim/uvsim.py
bhazelton: Yes, I think we should ditch the space and comma. I don't have a preference for '[' vs '<' vs '('. Do you?

mkolopanis: I don't have a preference myself.

bhazelton: done! I went with '<'.

pyuvsim/__init__.py
```diff
@@ -18,8 +18,8 @@
 except (LookupError, ImportError):
     try:
         # Set the version automatically from the package details.
-        __version__ = get_distribution(__name__).version
-    except DistributionNotFound:  # pragma: nocover
+        __version__ = version("pyuvsim")
```
jpober, line 21: Any concerns about this codecov warning?

pyuvsim/uvsim.py
```diff
@@ -522,7 +530,8 @@ def uvdata_to_task_iter(task_ids, input_uv, catalog, beam_list, beam_dict, Nsky_
     # some additional overhead for numpy arrays as well
     # usually seeing [0.00001, 0.00003] MiB / task memory required
     # based on the reference simulations.
-    order = np.lexsort((_bls[task_slice], _freqs[task_slice], _times[task_slice])).flat
+    # Flatten inside the loop so that we always have the same thing in each loop.
```
jpober, line 533: This comment is perhaps confusing.  Without knowing that this was the source of a bug, it seems weird to have something basically saying "Don't flatten here, flatten later".  Do we want to instead have a comment near line 564?

jpober (approved):
@bhazelton I'm fine with merging these in (don't want to delay on this any longer) but maybe take a quick peak at my two comments and see if you care to address either of them.

pyuvsim/uvsim.py
mkolopanis: I think I'll second this. Maybe just commend on like 564 to use flat so we have a 1-dimensional iterator. That was really the point of it to begin with since the iterable is originally an (nblts, freqs) array.

pyuvsim/__init__.py
bhazelton: The problem here is with the structure of the package and the fact that I removed the pragma that was preventing this line from being counted. We're talking in the wider RASG community about package structure, so for now I'd like to leave this alone.

pyuvsim/uvsim.py
bhazelton: will do!

jpober (approved):

PR #426, opened 2023-05-22
Fix MPI tests breaking on some builds

bhazelton:
## Description

The main problem we were seeing in some CI builds (and on my machine) seems to be related to not properly freeing shared memory resources. This adds code to keep track of the shared memory resources so they can be properly freed at the end of the simulation (or if there are errors raised).

Other minor fixes:
- Use up to date github actions.
- Fix tests for large nsky_parts that were failing locally for me.

## Motivation and Context

Some CI builds were breaking with MPI related errors.

## Types of changes

- [ ] Version change

## Checklist:

For all pull requests:

Bug fix checklist:

### DIFF
--- a/pyuvsim/tests/test_mpi.py
+++ b/pyuvsim/tests/test_mpi.py
@@ -88,6 +88,8 @@ def test_shared_mem():
     with pytest.raises(ValueError, match="assignment destination is read-only"):
         sA.itemset(0, 3.0)
 
+    mpi.world_comm.Barrier()
+
 
 def test_mem_usage():
     # Check that the mpi-enabled memory check is consistent
@@ -188,18 +190,19 @@ def test_big_bcast_gather_loop(fake_tasks):
 
 
 @pytest.mark.parallel(3)
-def test_sharedmem_bcast_with_quantities():
-    # Use mpi.quantity_shared_bcast and check returned objects.
+def test_sharedmem_bcast():
+    # Use mpi.shared_mem_bcast and check returned objects.
 
-    lats = Latitude(np.linspace(-np.pi / 2, np.pi / 2, 10), 'rad')
-    freqs = np.linspace(100e6, 130e6, 15) * units.Hz
-    lat_return = mpi.quantity_shared_bcast(lats)
-    freq_return = mpi.quantity_shared_bcast(freqs)
+    lats = np.linspace(-np.pi / 2, np.pi / 2, 10)
+    freqs = np.linspace(100e6, 130e6, 15)
+    lat_return = mpi.shared_mem_bcast(lats)
+    freq_return = mpi.shared_mem_bcast(freqs)
 
     if mpi.rank == 0:
         assert np.all(lat_return == lats)
         assert np.all(freq_return == freqs)
-        assert np.all(freq_return.to("MHz") == freqs.to("MHz"))
+
+    mpi.world_comm.Barrier()
 
 
 @pytest.mark.parallel(3)
--- a/pyuvsim/tests/test_uvsim.py
+++ b/pyuvsim/tests/test_uvsim.py
@@ -685,18 +685,21 @@ def test_set_nsky_parts_errors():
     mpi.start_mpi(block_nonroot_stdout=False)
 
     # Spoof environmental parameters.
-    # Choose an absurdly large number of tasks per node and very small available memory
-    # to get a large Nsky_parts_calc
-    # The alternative would be to make a very large source catalog, but that's not ideal in a test.
-    Npus_node = 2000
-    pyuvsim.mpi.Npus_node = Npus_node  # Absurdly large
-    os.environ['SLURM_MEM_PER_NODE'] = str(1000.)  # Only 10MB of memory
+    # Choose an absurdly large number of tasks per node and a lot of sources to get a
+    # large Nsky_parts_calc
+    Npus_node = 2000  # Absurdly large
+    mpi.Npus_node = Npus_node
+    mem_per_node = str(100000.)  # 1GB of memory per node
+    os.environ['SLURM_MEM_PER_NODE'] = mem_per_node
+    assert simutils.get_avail_memory() == float(mem_per_node) * 1e6  # convert to bytes
 
-    Nsrcs = 1e9
+    Nsrcs = 1e12
     cat_nfreqs = 1e3
 
     mem_avail = (simutils.get_avail_memory()
                  - mpi.get_max_node_rss(return_per_node=True) * 2**30)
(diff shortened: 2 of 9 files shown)


pyuvsim/tests/test_uvsim.py
```diff
@@ -685,18 +685,21 @@ def test_set_nsky_parts_errors():
...
-    # Choose an absurdly large number of tasks per node and very small available memory
+    # Choose an absurdly large number of tasks per node a lot of sources
     # to get a large Nsky_parts_calc
-    # The alternative would be to make a very large source catalog, but that's not ideal in a test.
-    Npus_node = 2000
-    pyuvsim.mpi.Npus_node = Npus_node  # Absurdly large
-    os.environ['SLURM_MEM_PER_NODE'] = str(1000.)  # Only 10MB of memory
+    Npus_node = 2000  # Absurdly large
+    mpi.Npus_node = Npus_node
+    mem_per_node = str(100000.)  # 1GB of memory per note
```
aelanman, line 692: in comment note->node

pyuvsim/tests/test_mpi.py
```diff
@@ -188,18 +191,21 @@ def test_big_bcast_gather_loop(fake_tasks):
 
 
 @pytest.mark.parallel(3)
-def test_sharedmem_bcast_with_quantities():
-    # Use mpi.quantity_shared_bcast and check returned objects.
+def test_sharedmem_bcast():
```
aelanman, line 194: The point of this test is to verify that the quantity_shared_bcast function works properly with astropy Quantities. Do we no longer need that feature?

pyuvsim/telescope.py
```diff
@@ -520,6 +524,19 @@ def _obj_to_str(self, beam_model):
 
         return path
 
+    def free_shared(self):
```
aelanman, line 527: Instead of duplicating this function, it might be cleaner to do one of two things:
1) Define this as a utility function in `mpi.py` that accepts an arbitrary object. Then for both this and SkyModelData, call `mpi.free_shared(self)` at the appropriate times.
2) Define a class in mpi.py that adds the shared_win_list attribute and free_shared function, and make both BeamList and SkyModelData inherit from it.

I'd be happy to do this if you like, but this is fine if you just want to get this fixed.

aelanman (commented):
Thank you so much for fixing this! And sorry for dropping the ball here.

My only concern is that you've removed the `quantity_shared_bcast` function. I guess if we're not using it then it's not an issue, but we should check that it's not used anywhere.

I've also left a couple suggestions to avoid duplicating a function, but it's not an issue if you want to keep it that way.

steven-murray:
@bhazelton overall the code looks good, however I'd suggest something similar to @aelanman in that I don't think the code that does the MPI sharing should be attached to any particular class that is being shared. To me, separation of concerns (or single responsibility principle) would dictate making a class/function in the `mpi` module that is able to share an arbitrary object, and also maintains an internal "global" window list that can be released. Having all the MPI stuff inside the `SkyModel` and `BeamList` classes makes them less usable by other code.

pyuvsim/tests/test_mpi.py
bhazelton: The only place that the `quantity_shared_bcast` function was called was in this test. It's not used in the code, I assume because of the conversion from using a SkyModel object to using a SkyModelData object? So we removed it and converted this test to testing the underlying `sharedmem_bcast` function.

pyuvsim/telescope.py
bhazelton: Yep, I think consolidating the code is a good idea. I think I'll take Steven's approach if I can make it work sensibly.

bhazelton:
I'm realizing that I don't understand how MPI knows which shared array is which. But it seems like it's tied to the objects, so I'm not sure if the sharing can be moved out of the objects without breaking things. The freeing can be moved into a function as suggested in @aelanman's point 1. But I think you're asking for something more @steven-murray , I think you want the sharing moved out of the object as well?

steven-murray:
@bhazelton yeah, I'd probably have to look at the details more closely to understand what's really possible. In my mind the ideal kind of situation would be an api like the following:

```python
from pyuvsim import mpi, SkyModel, BeamList

sky = SkyModel()
beams = BeamList()

mpi.share(sky, beams)
do_stuff(sky, beams)
mpi.free(sky, beams)
```

It might not be possible to do something like that -- an alternative might be to provide a key:

```python
from pyuvsim import mpi, SkyModel, BeamList

sky = SkyModel()
beams = BeamList()

mpi.share('sky_key', sky)
mpi.share('beam_key', beams)

do_stuff(sky, beams)
mpi.free('sky_key', 'beam_key')
```

One could imagine having an `mpi.free_all()` function as well.

bhazelton:
@aelanman @mkolopanis I think you two are best placed to tell us what is most likely to work here. I really don't understand how the different shared arrays are being kept track of in MPI...

aelanman:
Maybe this is bad form, but could you have a module-level list that gets appended to every time a new window is created, then register a cleanup function at exit?
e.g., in mpi.py:
```
_OPEN_WINDOWS = []

def shared_mem_bcast():
. . . 
    win = MPI.Win.Allocate_shared(nbytes, itemsize, comm=node_comm)
    buf, itemsize = win.Shared_query(0)
    sh_arr = np.ndarray(buffer=buf, dtype=dtype, shape=shape)
    _OPEN_WINDOWS.append(win)
. . .

def _window_cleaning():
   for win in _OPEN_WINDOWS:
        win.Free()

atexit.register(_window_cleaning)
```

This leaves the responsibility of tracking open MPI windows to the mpi module, so the BeamList and SkyModelData classes don't need any modification.

steven-murray:
@aelanman's suggestion is approximately what I was thinking of. The only catch would be if different lists of windows are required to be free'd at different times (eg. if you had two simulations running or something like that?). Then you'd need a way to keep track of which sub-list belongs to which simulation. If all you need to do is make sure everything is freed before exiting Python (and don't care about exactly when, so long as it's after the computation is done), then @aelanman's suggestion works perfectly.

mkolopanis:
Yeah this would definitely be another way to do it too. I don't suspect we have too large a user base that runs simulations over and over in like a jupyter notebook or something like that. I'm personally more prone to cleaning up after yourself when you're done using the resources but either way should work fine.

bhazelton:
Ok, we fixed this as a group on the telecon to move the shared window tracking to the `mpi` module and free it from there with an `atexit` cleanup call.

steven-murray (approved):
Looks good to me, thanks!

PR #435, opened 2023-09-29
Adding Lunar simulation capability

### LINKED ISSUE #318: Demonstration of lunar simulation capabilities
We have the functionality through the lunarsky module but we haven't actually run a simulation and done a critical analyses of the resulting visibilities to check it.  Do they have reasonable fringe rates?  Can we image them?  Etc.

### LINKED ISSUE #408: uvw errors for non-terrestrial telescopes?
In our telecon the other day, we looked at the calculation of uvw's (which eventually go into the visibility calculation) for telescopes on the moon.  When initializing from a parameter file, the UVData method `set_uvws_from_antenna_positions` gets called, which I thought we decided calls code from utils.pyx (i.e. cython code) where gps parameters about the Earth curvature and such are hardcoded.  Taking another look, neither @aelanman nor I couldn't find where that method called the cython code.  @mkolopanis Can you remember where the issue was?  Am I misremembering which method called the cython with hardcoded earth parameters?

wps2n:
## Description
- Adds integration for the lunar mcmf frame in simsetup.py allowing pyuvsim to simulate telescopes on the lunar surface. 
- Adds tests that ensure things are working end-to-end.
- Require lunarsky>=0.2.2 and pyuvdata>=2.4.3, which led to many other dependency version updates.
- Removed a bunch of code associated with older pyuvdata versions that is no longer needed. Biggest change here is removing the `uvb_params` concept in `BeamList` objects.

## Motivation and Context
Well, we want to be able to simulate telescopes on the moon too...
closes #408
closes #318

## Types of changes

- [ ] Version change

## Checklist:

For all pull requests:
- [ ] I have read the [contribution guide](CONTRIBUTING.md).
- [ ] My code follows the code style of this project.

New feature checklist:


CHANGELOG.md
```diff
@@ -2,6 +2,9 @@
 
 ## [Unreleased]
 
+### Changed
+- Require pyuvdata >= 2.2.12 to properly support lunar sims.
```
bhazelton, line 6: Need to add the other, more substantial, changes.

pyuvsim/simsetup.py
```diff
@@ -1719,6 +1706,12 @@ def initialize_uvdata_from_params(
 
     uv_obj = UVData()
     uv_obj._set_future_array_shapes()
+
+    # Setting the frame and acceptable range for the moon
+    uv_obj._telescope_location.frame = uvparam_dict['telescope_frame']
+    if uv_obj._telescope_location.frame == 'mcmf':
+        uv_obj._telescope_location.acceptable_range = (1.71e6, 1.75e6)
```
bhazelton, line 1713: I don't think these lines are necessary any more, I think this is done by pyuvdata when you set the frame.

setup.py
```diff
@@ -52,7 +52,7 @@
         'numpy>=1.20',
         'psutil',
         'pyradiosky>=0.2.0',
-        'pyuvdata>=2.2.10',
+        'pyuvdata>=2.2.12',
```
bhazelton, line 55: I think we should consider upgrading to require v2.4.1 so that if the sims are saved out in files, the frame information is preserved.

pyuvsim/simsetup.py
```diff
@@ -1752,13 +1745,13 @@ def initialize_uvdata_from_params(
...
 
-        _set_lsts_on_uvdata(uv_obj)
+        uv_obj.set_lsts_from_time_array()
         # set the apparent coordinates on the object
         uv_obj._set_app_coords_helper()
 
     else:
         uv_obj._set_drift()
-        _set_lsts_on_uvdata(uv_obj)
+        uv_obj.set_lsts_from_time_array()
```
bhazelton, line 1754: These lines are not covered by tests.

bhazelton: Oh! that's because this is handling for older pyuvdata versions. I'll fix them.

bhazelton:
I don't understand the codecov report. It seems to have the wrong version of the code files or something.

.readthedocs.yml
```diff
@@ -15,6 +15,8 @@ build:
 sphinx:
   builder: html
   configuration: docs/conf.py
+# might want to do this in future:
+  # fail_on_warning: true
```
mkolopanis, line 19: Do we want to do this? Delete these lines? Or keep it around for "future" still?

pyuvsim/telescope.py
```diff
@@ -114,7 +110,6 @@ class BeamList:
     def __init__(
         self,
         beam_list=None,
-        uvb_params=None,
```
mkolopanis, line 117: I see that we're free ourselves from a workaround before but could this be being used by users too?
Should be be concerned with the API change and push for a new release?

pyuvsim/telescope.py
```diff
@@ -151,14 +136,7 @@ def __init__(
...
 
-        # If any UVBeam objects are passed in, update uvb_params:
-        list_uvb_params = {}
-        if uvb_params is None:
-            uvb_params = {}
-            if not self.string_mode:
-                list_uvb_params = self._scrape_uvb_params(self._obj_beam_list, strict=False)
-        self.uvb_params.update(list_uvb_params)
-        self.uvb_params.update(uvb_params)    # Optional parameters for the UVBeam objects
+        # If any UVBeam objects are passed in, set uvb_read_kwargs:
```
mkolopanis, line 139: doesn't even look like we're checking for UVBeam objects at all. just the read_kwargs. Do we need this comment still?

pyuvsim/tests/test_simsetup.py
```diff
@@ -677,7 +677,18 @@ def test_param_reader_errors(subdict, error, msg):
...
+    ("world", "selenoid"),
+    [
+        (None, None),
+        ("moon", "SPHERE"),
+        ("moon", "GSFC"),
+        ("moon", "GRAIL23"),
+        ("moon", "CE-1-LAM-GEO"),
+        ("moon", None),
+    ]
+)
```
mkolopanis, line 690: shouldn't we cover the world = "earth" case too?

pyuvsim/tests/test_telescope.py
```diff
@@ -209,24 +185,6 @@ def test_comparison(beam_objs):
     assert beamlist == beamlist2
 
 
-def test_no_overwrite(beam_objs):
-    if not hasattr(beam_objs[0], "_freq_interp_kind"):
-        # this can go away when we require pyuvdata version >= 2.4.2
-        pytest.skip()
```
mkolopanis, line 215: so we don't need this test at all? or just didn't need to make this check but should still keep the test?

mkolopanis (commented):
I just have a few minor questions. Looks pretty good though.

.readthedocs.yml
bhazelton: Hmm, I'm not sure.  I think the reason it's the only change in that file is because there were other changes that were not different after a rebase. I can take out this change.

pyuvsim/telescope.py
bhazelton: I think this was only used under the hood for tracking extra items on UVBeams that are no longer included now that we require the most recent pyuvdata version. We do have external testing against `hera_sim` which passed. @steven-murray do you have any concerns here?

We are planning a new release pretty soon, but if we're concerned about this as an API change we should have a period of deprecation before removing it.

pyuvsim/telescope.py
bhazelton: probably not.

pyuvsim/tests/test_simsetup.py
bhazelton: sure. I think None is equivalent to "earth" but we should make sure of that.

pyuvsim/tests/test_telescope.py
bhazelton: The functionality it's associated with has been removed. It was testing handling of `uvb_params`.

pyuvsim/telescope.py
bhazelton: Currently all the code related to `uvb_params` is unreachable unless you add weird things to it by hand.

mkolopanis: okay that's fine I think :+1:

mkolopanis (approved):
looks good thank you!

PR #449, opened 2024-02-25
test cleanup and improvements

### LINKED ISSUE #311: Test with no explicit test
There are a few tests in `test_run.py` which run a small simulation but do not seem to actually make any tests beyond "did it not fail?" Maybe we can investigate further to find something to actually assert against once it is finished?

https://github.com/RadioAstronomySoftwareGroup/pyuvsim/blob/94bd4713430cd3231e0b1bd4bbfc5a70299e7768/pyuvsim/tests/test_run.py#L139-L150

### LINKED ISSUE #314: Atomize unit tests
There are several tests, many probably my fault, that do several `assert`s. Ideally, each test should only really check one thing. Parametrizing and using fixtures can avoid excessive code duplication or repeating large operations.

bhazelton:
## Description

General test cleanup, making sure all the tests actually assert something useful and are atomized as much as possible. Also added some coverage for some missing lines and re-wrote some old test files that had lots of weird issues.

## Motivation and Context

fixes #311
fixes #314

## Types of changes

- [ ] Version change
- [x] Other: test cleanup

## Checklist:

For all pull requests:


pyuvsim/simsetup.py
```diff
@@ -1218,12 +1218,13 @@ def parse_telescope_params(tele_params, config_path='', freq_range=None, force_b
     if 'array_layout' not in tele_params:
         raise KeyError('array_layout must be provided.')
     array_layout = tele_params.pop('array_layout')
+    assert isinstance(array_layout, (str, dict)), "array_layout must be a string or a dict"
```
mkolopanis, line 1221: I feel like with how much other manual error raising, it might be more meaningful to raise something specific and not just use an assert. But I can be convinced.

pyuvsim/tests/test_antenna.py
```diff
@@ -47,7 +48,38 @@ def test_jones_set_spline(cst_beam, hera_loc):
...
+
+    # These are just values from a run, so this just tests for unexpected changes.
+    expected_jones = np.array(
+        [
+            [
+                [
+                    -4.57061296e-04 - 3.88626249e-04j,
+                    -7.28285993e-05 - 1.45479743e-04j,
+                    -7.28285993e-05 - 1.45479743e-04j
+                ], [
```
mkolopanis, line 63: is it just me, or would I have expected black to put these two '],' and '[' on different lines? How weird.

pyuvsim/tests/test_simsetup.py
```diff
@@ -704,125 +807,112 @@ def test_freq_parser_errors(freq_dict, msg):
...
+        ['start_time', 'end_time', 'integration_time'],
+        ['start_time', 'integration_time', 'Ntimes', 'duration_hours'],
+        ['end_time', 'integration_time', 'Ntimes', 'duration_hours'],
+        ['start_time', 'integration_time', 'Ntimes'],
+        ['end_time', 'integration_time', 'Ntimes'],
+        ['start_time', 'duration_hours', 'Ntimes'],
+        ['end_time', 'duration_hours', 'Ntimes'],
+        ['start_time', 'duration_hours', 'integration_time'],
+        ['end_time', 'duration_hours', 'integration_time'],
+        ['time_array'],
```
mkolopanis, line 823: It's been a bit since I've looked at the parse time params, want to make sure that these 11 cases are all we need and not all 32 from choosing each of the 5 options here.

mkolopanis (commented):
Thanks for putting in all this hard work. I have really just one question, but otherwise I think this is great.

pyuvsim/simsetup.py
bhazelton: sure, I can add that. I actually added this when I was trying to figure out a test failure and I fixed the thing that had caused this upstream. But I think you could hit this with an error in the param yaml file, so it is a bit user facing.

pyuvsim/tests/test_antenna.py
bhazelton: We haven't blackified this repo yet (it's coming after the lunar sim branch is merged), so I did this by hand. It'll probably change when we use black.

pyuvsim/simsetup.py
mkolopanis: An error in the param yaml reader would be the best place I agree

pyuvsim/tests/test_antenna.py
mkolopanis: OH! that'll do it! carry on.

mkolopanis (approved):
Thanks a bunch!

PR #453, opened 2024-03-21
remove  troublesome nsky_parts test

mkolopanis:
removes the old test where we try to force nsky_parts.

We have a new test where we manually set it and we can check that the proper output is printed from that test.


bhazelton (approved):
Yay!! I hate that test. Thank you!!

PR #465, opened 2024-05-13
Fix compatibility with pyuvdata v3.0, use UVData.new

### LINKED ISSUE #334: Allow conjugation convention specification in obsparam
should we add the option to specify conjugation convention in an obsparam? This could be important if a user specifies bls as baseline numbers and some conventions will have different numbers

### LINKED ISSUE #427: `initialize_uvdata_from_params` always returns a UVData with `integration_time=1` if `Ntimes=1`
As in the title, `initialize_uvdata_from_params` always returns a UVData with `integration_time=1` if `Ntimes=1`. `pyuvsim` version used is `1.2.5`

bhazelton:
## Description

Fix some things that will break in pyuvdata v3.0, particularly related to the new handling of telescope metadata.

This includes moving to using `UVData.new` for initializing UVData objects, which changes the default conjugation convention from `"ant2<ant1"` to `"ant1<ant2"`. So I added an option to set the conjugation convention in the parameter yaml files, which was requested in #334. This also naturally fixes #427.

This also fixes the version handling for editable installs. This was missed when we went to the src layout.

Fixes #427
Fixes #334

## Types of changes

- [ ] Version change

## Checklist:

For all pull requests:

Bug fix checklist:

New feature checklist:


bhazelton:
The warnings tests are failing with an astropy future warning on pyuvdata code, which will be fixed in pyuvdata v3.0.

src/pyuvsim/simsetup.py
```diff
@@ -1911,129 +2014,108 @@ def initialize_uvdata_from_params(
...
+            cat_name = "unprojected"
     elif "object_name" in param_dict:
-        uvparam_dict[uvparam_name] = param_dict["object_name"]
+        cat_name = param_dict["object_name"]
     else:
-        uvparam_dict[uvparam_name] = param_dict["cat_name"]
+        cat_name = param_dict["cat_name"]
+    phase_center_catalog = {0: {"cat_name": cat_name, "cat_type": "unprojected"}}
 
     uv_obj = UVData()
```
mkolopanis, line 2048: I think we can remove this line. We call UVData.new() in each block of this if statement.

src/pyuvsim/simsetup.py
```diff
@@ -2178,6 +2262,14 @@ def initialize_uvdata_from_keywords(
         List of polarization strings (or ints) to insert into object
     no_autos : bool
         If True, eliminate all auto correlations
+    conjugation_convention : str, optional
+        A convention for the directions of the baselines, options are:
+        'ant1<ant2', 'ant2<ant1', 'u<0', 'u>0', 'v<0', 'v>0'
```
mkolopanis, line 2267: where's my `w<0` convention :cry:

src/pyuvsim/simsetup.py
```diff
@@ -2286,11 +2378,17 @@ def initialize_uvdata_from_keywords(
...
+    }
+
     freq_params = {k: v for k, v in freq_params.items() if v is not None}
     time_params = {k: v for k, v in time_params.items() if v is not None}
     selection_params = {k: v for k, v in selection_params.items() if v is not None}
     tele_params = {k: v for k, v in tele_params.items() if v is not None}
     layout_params = {k: v for k, v in layout_params.items() if v is not None}
+    ordering_params = {k: v for k, v in ordering_params.items() if v is not None}
 
     uv_obj = UVData()
```
mkolopanis, line 2393: Is this particular instance of a UVdata object _only_ used in 2395? If so maybe we can just put the initializer into that list comprehension? I personally got a little confused why were were creating a blank object instead of using new, util i realized it was just for the extra_kwds dict below.

tests/test_simsetup.py
```diff
@@ -1081,7 +1089,48 @@ def test_param_select_redundant():
...
+        warn_type = None
+        warn_str = ""
+        param_dict2["ordering"] = order_dict
+    with uvtest.check_warnings(warn_type, match=warn_str):
+        uv_obj2 = pyuvsim.initialize_uvdata_from_params(param_dict2, return_beams=False)
+
+    if order_dict is not None:
+        uv_obj2.reorder_blts(
+            order="time", minor_order="baseline", conj_convention="ant1<ant2"
+        )
```
mkolopanis, line 1129: Is it prudent to check we are getting the correct ordering and convention before turning it back to match the orig object? Or do we do that somewhere else too?

mkolopanis (commented):
I've got a few minor comments honestly, but otherwise this looks really good.

src/pyuvsim/simsetup.py
```diff
@@ -1324,8 +1340,9 @@ def parse_telescope_params(
 
     # fill in outputs with just array info
     return_dict = {}
-    beam_list = BeamList([])
-    beam_dict = {}
+    if return_beams:
+        beam_list = BeamList([])
+        beam_dict = {}
```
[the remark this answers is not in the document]
bhazelton: It's needed for the early return if `return_beams` is set, but I think it's better to just move it down to that return so I did that.

src/pyuvsim/simsetup.py
bhazelton: I use it to detect whether the UVData object has a "telescope" object attached to it (the hasattr call in the next line).

mkolopanis: oh yeah i do see that. Missed it before. Yeah I guess I'm not a fan of binding variables we're not using.

tests/test_simsetup.py
bhazelton: yes. I did it and realized that the logic wasn't quite right. Thanks!

src/pyuvsim/simsetup.py
```diff
@@ -1911,129 +2012,109 @@ def initialize_uvdata_from_params(
...
+            "x_orientation": "x_orientation",
+            "instrument": "instrument",
+        }
+        for key, tel_key in telescope_param_map.items():
+            if key in param_dict:
+                tel_init_params[tel_key] = param_dict[key]
+            if key in tele_params:
+                tel_init_params[tel_key] = tele_params[key]
+        if "instrument" not in tel_init_params:
+            tel_init_params["instrument"] = tel_init_params["name"]
```
steven-murray, line 2065: I would actually try to move to a more native configuration format that collected the telescope parameters under a heading properly, like

```yaml
telescope:
    name: hera
    antenna_names: ...
    etc.
```

It feels natural to me to be able to do `tele = Telescope.from_yaml(yamlfile)`, where the yaml file is self-contained.

src/pyuvsim/simsetup.py
```diff
@@ -1911,129 +2012,109 @@ def initialize_uvdata_from_params(
...
+
+        uv_obj = UVData.new(
+            telescope=Telescope.new(**tel_init_params),
+            phase_center_catalog=phase_center_catalog,
+            vis_units="Jy",
+            history="",
+            do_blt_outer=True,
+            time_axis_faster_than_bls=True,
```
steven-murray, line 2073: Isn't this the wrong way around? You want the blt order to be `(time, baseline)` right? That would make time the outer (or slower) axis.

src/pyuvsim/simsetup.py
```diff
@@ -1911,129 +2012,109 @@ def initialize_uvdata_from_params(
...
-    # so order this now so we don't get any warnings.
-    if reorder_blt_kw is None:
-        reorder_blt_kw = {"order": "time", "minor_order": "baseline"}
-
-    if reorder_blt_kw and reorder_blt_kw != {
-        "order": uv_obj.blt_order[0],
-        "minor_order": uv_obj.blt_order[1],
-    }:
-        uv_obj.reorder_blts(**reorder_blt_kw)
+    ordering(uv_obj, param_dict, reorder_blt_kw)
```
steven-murray, line 2113: this feels weird... can the name of the function be a verb like `reorder_blts` or `set_order`?

steven-murray (commented):
Thanks @bhazelton, just a few smallish comments from me

src/pyuvsim/simsetup.py
```diff
@@ -1754,34 +1785,82 @@ def subselect(uv_obj, param_dict):
...
+    bl_conjugation_convention = None
+
+    # this is the order required by the simulator. Might as well default to it.
+    default_blt_order = ["time", "baseline"]
+    if "ordering" in param_dict:
+        ordering_dict = param_dict["ordering"]
+        if "blt_order" not in ordering_dict:
+            ordering_dict["blt_order"] = default_blt_order
+    else:
+        ordering_dict = {"blt_order": default_blt_order}
```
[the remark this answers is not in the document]
bhazelton: You're correct, but I needed to set a default and this is what is actually enforced inside the simulator. Als, if you call `reorder_blts` and only specify time first, you'll get `["time", "baseline"]`, so it seemed a reasonable default.

src/pyuvsim/simsetup.py
bhazelton: oops, you're right, thanks!

src/pyuvsim/simsetup.py
bhazelton: sure.

src/pyuvsim/simsetup.py
bhazelton: I can look at that, but I was trying not to make changes to the pyuvsim API.

src/pyuvsim/simsetup.py
bhazelton: I actually made the corresponding mistake in the pyuvdata telescope_refactor PR as well. I'll fix it there too.

src/pyuvsim/simsetup.py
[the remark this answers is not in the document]
bhazelton: I suppose with the new `time_axis_faster_than_bls` and `blts_are_rectangular` attributes I can change the simulator to just check for those instead, I'll play with that.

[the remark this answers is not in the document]
mkolopanis: the time,baseline requirement was just to minimize re-compute on beams and coherency right?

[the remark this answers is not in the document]
steven-murray: ah, if there's a good reason for it to be (time, baseline), then by all means. And I have no problem with having this as a default at all. I thought I saw that you were changing the ordering to (time, baseline) even if the user requested otherwise, and then simply changing the ordering back at the end. If so, that could be some time wasted if (time, baseline) is not strictly required and the user still requested something rectangular (like (time, ant1)).

src/pyuvsim/simsetup.py
steven-murray: Yeah, that's a fair concern, of course. Probably both could be supported for some time. Maybe this is something to bring up in a separate issue.

src/pyuvsim/simsetup.py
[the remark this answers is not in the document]
bhazelton: I actually think that the baseline order within a time does matter for computational efficiency because of the beam handling. So I'm not inclined to change what is done inside the simulator.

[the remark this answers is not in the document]
steven-murray: OK, no problem

steven-murray (approved):

PR #476, opened 2024-07-12
Fixed a bug in run_uvdata_uvsim to ensure correct uvws

### LINKED ISSUE #473: pyuvsim ignoring (and not updating) `UVData.uvw_array`
I stumbled upon an interesting feature when helping out a user who was running into an issue taking a `UVData` object from a prior observation, passing it through pyuvsim, and then writing out the result as an MS file (via the `UVData.write_ms` method). Long story short, what we found was that when passing a `UVData` object, `run_uvdata_uvsim` does not use the `uvw_array` values for calculating the interferometer response, favoring instead using the antenna positions to internally generate ENUs (and then calculating fringe response from there). This would be fine, except that `run_uvdata_uvsim` does _not_ appear to update `uvw_array` when it returns the object, which means that an object with incorrect uvw coordinates gets passed back.

At a minimum it would be helpful to raise an error when the values in `uvw_array` are not consistent w/ the antenna position-based values, though I think a better solution would be to simply overwrite the `uvw_array` values and wipe out any info on the phase centers (though I don't know the downstream ramifications of this, and figure this is a question for a more experienced dev/user to address). Happy to a write up a PR if someone wants to recommend which path to take.

bhazelton:
## Description

Fixes a bug which resulted in incorrect uvws for output UVData object.

## Motivation and Context

Fixes #473

## Types of changes

- [ ] Version change

## Checklist:

For all pull requests:

Bug fix checklist:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,6 +15,10 @@ changes for numpy 2.0 compatibility.
 - Use UVData.new method to do the UVData object initialization. This leads to a
 change in the default conjugation convention from `"ant2<ant1"` to `"ant1<ant2"`.
 
+### Fixed
+- Fixed a bug where `run_uvdata_uvsim` does not update uvw_array to what it actually
+used for the simulation, resulting in output objects with incorrect uvw arrays.
+
 ### Deprecated
 - The `"reorder_blt_kw"` parameter in the `simsetup.initialize_uvdata_from_params`
 function in favor of setting the ordering via the new "ordering" section in the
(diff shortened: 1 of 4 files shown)


mkolopanis (approved):
works for me. Good idea moving some of that finalization into a new function

bhazelton:
> works for me. Good idea moving some of that finalization into a new function

That refactoring was required because my added `if` statement made it too complex 😝

PR #487, opened 2024-08-22
Use pyuvdata Analytic Beams and the new BeamInterface class

### LINKED ISSUE #443: Use the new pyuvdata Analytic Beam and Beam Interface objects
To replace the current Analytic Beams in this package.

bhazelton:
## Description

Use the newly refactored Analytic Beams in pyuvdata and the new BeamInterface object that unifies the interface for interpolation of UVBeams & evaluation of Analytic Beams.

Note: This is a significant change for "unpolarized" Analytic beams, including uniform, gaussian and Airy beams. The previous implementation in pyuvsim actually implemented these as polarized beams, with one polarization sensitive only to the zenith angle oriented emission and the other sensitive only to azimuthally oriented emission. The new implementation in pyuvdata makes these actually unpolarized so that the two instrumental polarizations are equally sensitive to both zenith angle and azimuthally oriented polarization and **are identical to each other.** This last point leads to the biggest change -- the instrumental cross pol ("xy" and "yx") visibilities for these unpolarized beams are equal to the same pol ("xx" and "yy") visibilities, whereas with the previous implementations they were zero for unpolarized sources. In particular, this will lead to differences in reference simulations that use these kinds of beams.

Related changes:
- Remove all analytic beam specifications in pyuvsim.
- BeamLists now contain BeamInterface objects (but are still initialized with lists of UVBeam and/or Analytic Beam objects)
- Significant refactoring of BeamLists, removing the string/object mode and replacing it with a shared memory approach similar to what is done for SkyModel objects.
- require at least pyuvdata v3.1.0.
- Remove handling for older pyuvdata versions.

I also bumped the version to 1.4.0 as this is a pretty major change and removed code that was marked to be removed in version 1.4.

## Motivation and Context

closes #443

## Types of changes

- [x] Breaking change (fix or feature that would cause existing functionality to change)
- [x] Version change

## Checklist:

For all pull requests:

New feature checklist:

This does change the reference sims because the analytic unpolarized beams are now truly unpolarized.


bhazelton:
Two parallelized tests are currently timing out. I'm worried that's related to how I'm handling the shared memory stuff, would love someone who's more familiar with MPI to take a look. Maybe @mkolopanis ?

src/pyuvsim/uvsim.py
```diff
@@ -812,9 +785,12 @@ def run_uvdata_uvsim(
     Npus = mpi.get_Npus()
 
     input_uv = comm.bcast(input_uv, root=0)
-    beam_list = comm.bcast(beam_list, root=0)
     beam_dict = comm.bcast(beam_dict, root=0)
     catalog.share(root=0)
+    if rank == 0:
+        print(beam_list)
+        print(type(beam_list))
```
steven-murray, line 792: is this meant to be here?

steven-murray (commented):
Thanks @bhazelton! I think this all looks good -- honestly can't say too much about the MPI stuff without looking at it in more detail. The one thing I think that can be better is the simsetup around the beam, but I think that should go input pyuvdata.

src/pyuvsim/data/test_config/28m_triangle_10time_10chan_nodiameter.yaml
```diff
@@ -1,12 +1,13 @@
 beam_paths:
-  0:
+  0: !UVBeam
     filename: HERA_NicCST.beamfits
-  1:
-    type: uniform
-  2:
-    type: gaussian
+    path_variable: pyuvsim.data.DATA_PATH
```
steven-murray, line 4: I see how you're using the `path_variable` here (cf my question on `pyuvdata`). It's possible we could use entry-points to define "plugins" to pyuvdata that add to a global "BEAM_PATHS" list, which could be searched through for filenames, in which case we wouldn't need this variable at all. Lemme know if you're interested in that.

bhazelton: Interesting. I really think this is primarily needed for testing, I don't expect it to really be used by users much at all, but I needed to make it work for the pyuvsim tests (I wasn't planning to advertise it in the docs). If you think the plugin structure is better we can explore it, but since this is really just for testing I don't know that we need to put a lot of effort into it.

steven-murray: Yeah, OK let's keep as-is for now. I think it's clear enough.

src/pyuvsim/uvsim.py
bhazelton: no, thanks!

src/pyuvsim/telescope.py
```diff
@@ -329,219 +283,99 @@ def beam_type(self):
...
-            self._obj_beam_list[ind] = value
+    def __setitem__(self, ind, value):
+        """Insert into the beam list."""
+        try:
+            self.beam_list[ind] = value
+            # reset the is_consistent flag when changing a value
+            self.is_consistent = False
+        except ValueError as err:
+            self.beam_list.pop()
+            raise err
```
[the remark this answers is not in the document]
bhazelton: I think that's reasonable. Do you think we need any deprecation on going to an immutable list? We're changing a lot of other things about BeamLists in this PR, so maybe it's ok to just change this too.

I can remove the `__setitem__` and `append` methods, but I'm not sure if that is enough to truly make it immutable. Thoughts?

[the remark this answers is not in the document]
bhazelton: I think I figured it out.

src/pyuvsim/telescope.py
```diff
@@ -61,112 +60,69 @@ class BeamList:
...
-    _float_params = {
-        "sig": "sigma",
-        "diam": "diameter",
-        "reff": "ref_freq",
-        "ind": "spectral_index",
-    }
+    beams: InitVar([UVBeam | AnalyticBeam])
+    _: KW_ONLY
+    beam_type: Literal["efield", "power"] | None = "efield"
+    spline_interp_opts: dict[str:int] | None = None
```
steven-murray, line 83: I thought the specification of a dict type was in the format `dict[keytype, valtype]` (i.e. comma instead of colon)?

src/pyuvsim/telescope.py
```diff
@@ -61,112 +60,69 @@ class BeamList:
...
+            be True for the pyuvsim simulator. Defaults to True.
 
-        self.uvb_read_kwargs = uvb_read_kwargs or {}
-        self.select_params = select_params or {}
-        self.is_consistent = False
-        if check and (not self.string_mode or force_check):
-            # only call the check if it will do something. This avoids an annoying warning.
-            self.check_consistency(force=force_check)
+        """
+        beam_list = []
```
steven-murray, line 101: I wonder if using a tuple instead of a list isn't better? I also wonder if it would be easier to have the input as `beams` and the attribute saved onto the instance also called `beams`, so you don't have to go through these shenanigans of having a separate init=False variable etc. That is, think of the `__post_init__` as simply doing a conversion from the input `beams` to a `beams` attribute where each beam is simply normalized properly, and the list is converted to tuple.

steven-murray (commented):
Only a couple more comments from me -- pending the uvdata PR being finalized, and tests working, I think this is all good now.

src/pyuvsim/telescope.py
bhazelton: I used to have the two variables have the same name (both `beam_list`), I just worried it might be confusing. I can change it back.

I went and read a lot about tuples vs lists, e.g. the top rated answer to this StackOverflow question and references therein: https://stackoverflow.com/questions/626759/whats-the-difference-between-lists-and-tuples

After reading those references, I decided list was maybe more appropriate, and since I'm using the `frozen` keyword to dataclass, the immutability difference isn't an issue. But I'm willing to go to a tuple if you think it makes more sense.

steven-murray (approved):
THanks @bhazelton this all looks good now -- though I'm a bit worried about that failing test that seems to reveal an MPI problem. @mkolopanis any thoughts?

PR #503, opened 2024-10-21
perf: allow not running checks

steven-murray:
## Description

Since `UVData.new()` already runs an object check, we pass the `check_kw` to it instead of running the object check in `_complete_data`. The object check actually takes a somewhat significant amount of time (because it has to recompute all the LSTs to check them...) so it's good to give the user control over when it should be run, and remove duplication.

## Motivation and Context

HERA validation sims.

## Types of changes

- [x] Performance enhancement
- [ ] Version change

## Checklist:

For all pull requests:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,7 +9,7 @@
 - Major restructuring of the BeamList object to use shared memory and remove the
 string/object mode switch.
 - BeamLists now contain pyuvdata BeamInterface objects
-- Updated minimum dependency versions: pyuvdata>=3.1.0
+- Updated minimum dependency versions: pyuvdata>=3.1.2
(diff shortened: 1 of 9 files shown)


bhazelton:
Does this require some of the recent changes to pyuvdata?

steven-murray:
@bhazelton yes it requires the `check_kw` parameter to `UVData.new()` which came in 3.1.2.

steven-murray:
Are we OK requiring >=3.1.2, or do you want me to do a version check in the code?

bhazelton (commented):
need to also update `environment.yml` and the readme with these new versions.

bhazelton (approved):
Looks good.

PR #511, opened 2024-11-22
Add windows support

steven-murray:
## Description

This adds `psutil` to the dependencies for the `[sim]` extra, and moves the import of `resource` / `psutil` into the only function that uses it. This means that Windows can now run sims.

Note that some tests still fail on windows (e.g. using miriad files). I haven't dug deep into it, but we could try putting xfails on thins we know will fail and start testing against windows if we want. For now, this update means all matvis tests pass on windows.

## Motivation and Context

In testing matvis against pyuvsim, I was trying to use the GPU, and on my laptop that's easier to do on Windows. I found that the pyuvsim comparison tests broke on Windows however, simply because of the import of `resource`, which is unix-only. This fixes it.

## Types of changes

- [ ] Version change

## Checklist:

For all pull requests:


bhazelton:
This is great! The myriad part of pyuvdata doesn't work on Windows, so that's why those break.

bhazelton:
I suspect we should update the readme and maybe some other docs to mention windows compatibility.

.github/workflows/windows-tests.yaml
```diff
@@ -0,0 +1,65 @@
...
+    runs-on: windows-latest
+    defaults:
+      run:
+        # Adding -l {0} helps ensure conda can be found properly.
+        shell: bash -l {0}
+    strategy:
+      fail-fast: false
+      matrix:
+        python-version: ["3.10", "3.11", "3.12"]
+        env_name: [pyuvsim_tests_windows]
```
bhazelton, line 24: Unless there are differences beyond just the os and env_name, I think this could be added to the existing testsuite.yaml workflow using the `include` options, similar to what we did for mpich. That would make it easier to maintain.

pyproject.toml
```diff
@@ -49,13 +49,13 @@ classifiers = [
...
 moon = ["lunarsky>=0.2.5"]
-sim = ["mpi4py>=3.1.1"]
+sim = ["mpi4py>=3.1.1", "psutil"]
 all = ["pyuvsim[casa,healpix,moon,sim]"]
 test = ["coverage", "pre-commit", "pytest", "pytest-cov>=5.0"]
 doc = ["matplotlib", "pypandoc", "sphinx"]
 profiler = ["line-profiler"]
 dev = ["pyuvsim[all,test,doc,profiler]"]
-
+windows-dev = ["pyuvsim[test,doc,profiler,healpix,moon,sim]"]
```
bhazelton, line 58: I think we need to document this new option in the readme.

We also need some more general docs updates to make it clear that pyuvsim works on Windows.

tests/test_utils.py
```diff
@@ -153,9 +156,15 @@ def test_write_uvdata(save_format, tmpdir):
...
     """Test overwriting a uvdata object yields the expected results."""
+    if sys.platform.startswith("win"):
+        pytest.skip("Clobbering is flaky on windows")
+
     if save_format == "ms":
         pytest.importorskip("casacore")
 
+    if save_format == "miriad" and sys.platform.startswith("win"):
+        pytest.skip("miriad is not supported on windows")
+
```
bhazelton, line 167: I think this isn't needed because the skip three lines earlier will have already skipped this test on windows.

bhazelton (commented):
This is awesome @steven-murray, thank you! I think the main thing we still need are updates to the changelog, readme and other docs to make it clear to users that pyuvsim is now supported on Windows.

mkolopanis:
also excited at the prospect but I don't have any substantial comments beyond what Bryna has said already.

.github/workflows/windows-tests.yaml
steven-murray: The one difference is that for windows we don't use `-n auto` because it can result in weird filesystem errors.

tests/test_utils.py
steven-murray: Good catch.

pyproject.toml
steven-murray: OK, I added:

* Trove classifiers to the pyproject.toml to make it clear that windows is supported
* Several sentences to the readme (in difference sections) to make it clear that windows is (partially) supported, and instruct on how to install for a dev environment on windows.

.github/workflows/windows-tests.yaml
steven-murray: Ah yeah, sure!

bhazelton: Thanks!

bhazelton (approved):
Looks great, thanks @steven-murray!!

bhazelton (approved):

PR #528, opened 2025-02-27
Fix check_kw handling and changelog errors

bhazelton:
## Description

This PR fixes some errors introduced in recent PRs.

I discovered errors in how #503 was implemented, so I fixed them here.

I also discovered that there were mistakes in changelog updates in #503 and #519 where changes were recorded as going into version 1.4.0 rather than into the unreleased section. These are fixed here as well.

## Types of changes

- [ ] Version change

## Checklist:

For all pull requests:

Bug fix checklist:

### DIFF
--- a/src/pyuvsim/simsetup.py
+++ b/src/pyuvsim/simsetup.py
@@ -1902,7 +1902,10 @@ def initialize_uvdata_from_params(
         the setup process. Typical parameters include "order" and "minor_order". Default
         values are ``order='time'`` and ``minor_order='baseline'``.
     check_kw : dict (optional)
-        Deprecated and has no effect.
+        A dictionary of keyword arguments to pass to the :func:`uvdata.UVData.check`
+        method after object creation. Caution: turning off critical checks can
+        result in a UVData object that cannot be written to a file.
+
 
     Returns
     -------
@@ -1925,12 +1928,6 @@ def initialize_uvdata_from_params(
             DeprecationWarning,
         )
 
-    if check_kw is not None:
-        warnings.warn(
-            "The check_kw parameter is deprecated and has no effect. This will "
-            "become an error in version 1.5",
-            DeprecationWarning,
-        )
     uvparam_dict = {}  # Parameters that will go into UVData
     if isinstance(obs_params, str):
         param_dict = _config_str_to_dict(obs_params)  # Container for received settings.
@@ -2078,6 +2075,7 @@ def initialize_uvdata_from_params(
         do_blt_outer=True,
         time_axis_faster_than_bls=False,
         antpairs=antpairs,
+        check_kw=check_kw,
         **uvparam_dict,
     )
 
@@ -2104,12 +2102,11 @@ def initialize_uvdata_from_params(
         return uv_obj
 
 
-def _complete_uvdata(uv_in, inplace=False, check_kw=None):
+def _complete_uvdata(uv_in, inplace=False):
     """
-    Fill out all required parameters of a :class:`pyuvdata.UVData` object.
+    Initialize data-like arrays on a a :class:`pyuvdata.UVData` object.
 
-    Ensure that it passes the :func:`pyuvdata.UVData.check()`.
-    This will overwrite existing data in `uv_in`!
+    This will overwrite any existing data in `uv_in` unless inplace=True.
 
     Parameters
     ----------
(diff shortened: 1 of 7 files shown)


src/pyuvsim/simsetup.py
```diff
@@ -2145,13 +2139,6 @@ def _complete_uvdata(uv_in, inplace=False, check_kw=None):
     uv_obj.flag_array = np.zeros(_shape, dtype=bool)
     uv_obj.nsample_array = np.ones(_shape, dtype=float)
 
-    uv_obj.extra_keywords = {}
```
bhazelton, line 2148: I do not think we should be re-setting these because they actually contain useful info like what the input yaml files are.

src/pyuvsim/simsetup.py
```diff
@@ -2145,13 +2139,6 @@ def _complete_uvdata(uv_in, inplace=False, check_kw=None):
...
-
-    if check_kw is not False:
-        if check_kw is None:
-            check_kw = {}
-        uv_obj.check(**check_kw)
-
```
bhazelton, line 2154: I don't think there is any way that the things done in this function can cause checks to fail if the input object is ok. Since this is a private method, we don't need to guarantee behavior for random users, only ourselves, so I think this is unnecessary.

src/pyuvsim/simsetup.py
bhazelton: Note that this change led to a number of test changes but I think it's the right thing to do.

src/pyuvsim/simsetup.py
```diff
@@ -2078,6 +2075,7 @@ def initialize_uvdata_from_params(
         do_blt_outer=True,
         time_axis_faster_than_bls=False,
         antpairs=antpairs,
+        check_kw=check_kw,
```
bhazelton, line 2078: In #503 `initialize_uvdata_from_keywords` was updated to pass `check_kw` to this function (`initialize_uvdata_from_params`) with the justification that the UVData.new method can do the checks, so they don't need to be done in `_complete_uvdata`. However, we'd already marked `check_kw` in `initialize_uvdata_from_params` as deprecated and having no effect. So I revived it here and passed it into UVData.new(). However I could also support just removing it so that all checks are run when we make a new object.

steven-murray: I definitely do not want to have to do all checks when making a new object, because in practice we often split simulations into multiple parts, each with its own config YAML, and its overkill to check object creation for each, when each is just changing e.g. the frequency channel simulated. When each individual simulation is somewhat small (one channel) the setup time can be a non-negligible overhead.

steven-murray: So, keeping this argument somewhere is a good idea.

steven-murray (approved):
Thanks @bhazelton. This all looks good to me!

steven-murray (commented):
`hera_sim` tests are failing for two reasons. One of them is that they use the `check_kw` argument for `_complete_uvdata`, so I've made a PR to fix that. The other however doesn't seem to be to do with `hera_sim` itself -- it's an error that `telecfg` is not a key in `extra_keywords`, and the error seems to be coming from a function in pyuvsim itself, so we should check that out.

bhazelton:
> `hera_sim` tests are failing for two reasons. One of them is that they use the `check_kw` argument for `_complete_uvdata`, so I've made a PR to fix that. The other however doesn't seem to be to do with `hera_sim` itself -- it's an error that `telecfg` is not a key in `extra_keywords`, and the error seems to be coming from a function in pyuvsim itself, so we should check that out.

I fixed the extra keywords problem, but if hera_sim is calling the (private) `_complete_uvdata` method then maybe I should just put the check_kw option back in. I took it out because I figured only trusted code was calling it.

steven-murray:
> I fixed the extra keywords problem, but if hera_sim is calling the (private) `_complete_uvdata` method then maybe I should just put the check_kw option back in. I took it out because I figured only trusted code was calling it.

We understand that it's a private method, so API changes are allowed to happen. If it's made a public API, then more stability would be required, but I think it's fine. It's a simple fix on the hera-sim end.

bhazelton:
@steven-murray if you're good with this now can you re-approve?

steven-murray (approved):
Great!

PR #532, opened 2025-03-18
Only import lunarsky if it's actually needed

bhazelton:
## Description

Lunarsky does a bunch of calculations and caching on import. So we should only import it if it's actually needed.

## Types of changes

- [ ] Version change

## Checklist:

For all pull requests:

Bug fix checklist:

### DIFF
--- a/src/pyuvsim/simsetup.py
+++ b/src/pyuvsim/simsetup.py
@@ -12,6 +12,7 @@ objects and empty :class:`pyuvdata.UVData` objects from configuration files.
 """
 
 import ast
+import contextlib
 import copy
 import logging
 import os
@@ -69,13 +70,6 @@ except ImportError:
         return 0
 
     mpi = None
-
-try:
-    from lunarsky import LunarTopo, MoonLocation, SkyCoord as LunarSkyCoord
-
-    hasmoon = True
-except ImportError:
-    hasmoon = False
 
 
 from .utils import check_file_exists_and_increment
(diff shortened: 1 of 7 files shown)


src/pyuvsim/simsetup.py
```diff
@@ -2019,6 +2027,9 @@ def initialize_uvdata_from_params(
             *tele_params["telescope_location"], unit="m"
         )
     elif tele_params["telescope_frame"] == "mcmf":
+        # to get here, lunarsky has to be installed, so don't need to test for it
+        from lunarsky import MoonLocation
```
mkolopanis, line 2031: Not that I don't believe you in this comment, But i cannot actually find _where_ we have already checked for Lunarsky here. If we're making a mock catalog it gets called there but what about the case where we don't?

tests/conftest.py
```diff
@@ -266,7 +259,9 @@ def hera_loc():
 
 @pytest.fixture(scope="session")
 def apollo_loc():
-    if hasmoon:
+    try:
+        from lunarsky import MoonLocation
+
         return MoonLocation(lat=0.6875, lon=24.433, height=0)
-    else:
+    except ImportError:
```
mkolopanis, line 266: Is there any advantage to the contextlib supression vs a try/except? the rest of the code base for this change seems to like the context lib. Maybe this try/except just reads easier for you in this fixture?

mkolopanis (commented):
Just a few minor comments

tests/conftest.py
mkolopanis: Totally understandable, I asked because it was different from how the other import attempts were in this update.

mkolopanis (approved):

PR #533, opened 2025-03-21
Fixes to be compatible with new telescope attributes in pyuvdata

bhazelton:
## Description

This adds compatibility with the new feed_angle, feed_array, and mount_type parameters in pyuvdata's Telescope object and beam objects.

The key here is to use the feed information on the beams to properly fill out the corresponding info on the UVData.telescope object.

## Types of changes

- [ ] Version change

## Checklist:

For all pull requests:

New feature checklist:

### DIFF
--- a/tests/test_simsetup.py
+++ b/tests/test_simsetup.py
@@ -21,6 +21,7 @@ from pyuvdata import (
     UniformBeam,
     UVBeam,
     UVData,
+    utils as uvutils,
 )
 from pyuvdata.analytic_beam import AnalyticBeam
 from pyuvdata.data import DATA_PATH as UV_DATA_PATH
@@ -135,7 +136,7 @@ def cat_with_some_pols():
 
 @pytest.fixture
 def uvdata_keyword_dict():
-    return {
+    init_dict = {
         "antenna_layout_filepath": os.path.join(
             SIM_DATA_PATH, "test_config/triangle_bl_layout.csv"
         ),
@@ -158,6 +159,14 @@ def uvdata_keyword_dict():
         "write_files": False,
         "run_check": True,
     }
+
+    if hasattr(Telescope(), "feed_array"):
+        # always do this once we require pyuvdata >= 3.2
+        init_dict["feed_array"] = ["x", "y"]
+        init_dict["feed_angle"] = [np.pi / 2, 0]
+        init_dict["mount_type"] = "fixed"
+
+    return init_dict
 
 
 @pytest.mark.parametrize("return_data", [True, False])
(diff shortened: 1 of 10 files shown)


tests/test_simsetup.py
```diff
@@ -1054,6 +1115,7 @@ def test_param_select_cross():
     assert uv_obj_cross == uv_obj_cross2
 
 
+@pytest.mark.filterwarnings("ignore:No beam information, so cannot determine")
```
steven-murray, line 1118: Probably we should update our test data to include beam info so we don't have to ignore these warnings.

steven-murray (commented):
Thanks @bhazelton. This is all looking good, with one minor comment. Is the plan to release v3.2 before merging this?

tests/test_simsetup.py
bhazelton: It's not a problem with the test data, we're explicitly turning off getting beam information in the call below (`return_beams=False`). The option to do this was something we added at your request, actually. But it does speed things up and this test doesn't need that stuff to be set, it's testing unrelated functionality.

bhazelton:
> Thanks @bhazelton. This is all looking good, with one minor comment. Is the plan to release v3.2 before merging this?

I was thinking we could get this in before 3.2 is released because it is backwards compatible. But open to waiting.

tests/test_simsetup.py
steven-murray: Ahh I see. Still, this feels wrong. We don't want a thousand warnings when you're doing something you want to (and are allowed to) do. Is there a way to catch this kind of circumstance so we don't raise a warning?

kartographer:
@bhazelton -- this all looks good/consistent on my end with https://github.com/RadioAstronomySoftwareGroup/pyuvdata/pull/1521

tests/test_simsetup.py
bhazelton: I'm on the fence about this. I added this warning in this PR because for the first time there's a real consequence to the UVData metadata if you don't make the beams. Maybe the right thing to do is to add support for specifying this info in the telescope config file, in which case it can get it from there if (and only if) the beams aren't constructed?

steven-murray: I think this is the right thing to do. If there is necessary uvdata metadata that can only be gotten from the beams, then we have to make a pathway to specifying that info outside specifying the beams.

bhazelton: I started digging into this and I realized that with the new yaml constructors for the Analytic and UVBeam objects, the beams are basically always read in already, so there's not really a performance reason not to use the beams to set these parameters properly. And adding ways to pass them separately from the beams makes it possible for the information to get out of sync between the beams and the yaml file. So I think I'm going to change the code to use the beam information if it's available and only issue the warning if the beam info isn't available.

bhazelton: Ok, I ended up halfway between. If you call `uvdata_init_from_params`, it gets the feed information from the beams specified in the telescope config file. If you call `uvdata_init_from_keywords` there are now options to pass the `feed_array`, `feed_angle`, and `mount_type` into that function.

PR #537, opened 2025-04-02
Improve the way we're setting the phase center catalog name

### LINKED ISSUE #536: Bug in the name of the phase center when combining multiple files in UVData.from_file.
When reading and concatenating multiple files with `UVData.from_file()`, I sometimes have to set the keyword `ignore_name` to `True` to enable concatenation. Otherwise, an error arises:
> ValueError:  UVParameter phase_center_id_array does not match. Cannot combine objects.

 This error arises even if the `phase_center_id_array` is the same for each of my `uvh5` file.

bhazelton:
Makes concatenation easier, also makes more sense

## Description

The way we're setting the phase catalog center name right now doesn't make much sense and leads to errors when files are concatenated after simulation. These errors are surmountable with the `ignore_name` parameter, but the whole situation is a unnecessary and would go away with this fix.

## Motivation and Context

fixes #536 

## Types of changes

- [ ] Version change

## Checklist:

For all pull requests:

New feature checklist:

### DIFF
--- a/src/pyuvsim/simsetup.py
+++ b/src/pyuvsim/simsetup.py
@@ -2106,18 +2106,7 @@ def initialize_uvdata_from_params(
         telescope_location.ellipsoid = tele_params["ellipsoid"]
 
     if "cat_name" not in param_dict and "object_name" not in param_dict:
-        tloc = EarthLocation.from_geocentric(
-            *tele_params["telescope_location"], unit="m"
-        )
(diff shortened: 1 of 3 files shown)


kartographer (commented):
This all looks good to me -- I think you're just missing a CHANGELOG entry.

steven-murray (approved):
Great!

PR #540, opened 2025-04-15
Fix memory issue associated with SkyModels that are not "full" or "flat" spectral type

bhazelton:
## Description

Call `skymodel.at_frequencies` in the simulation setup phase rather than per compute node.

## Types of changes

- [ ] Version change

## Checklist:

For all pull requests:

Bug fix checklist:

### DIFF
--- a/src/pyuvsim/uvsim.py
+++ b/src/pyuvsim/uvsim.py
@@ -40,7 +40,7 @@ class UVTask:
 
     Parameters
     ----------
-    sources : :class:`pyuvsim.simsetup.SkyModelData`
+    sources : :class:`pyradiosky.SkyModel`
         The sources to include in the visibility.
     time : :class:`astropy.time.Time` object or float
         Time at which to calculate the visibility, either an astropy Time object or a
@@ -59,7 +59,7 @@ class UVTask:
 
     Attributes
     ----------
-    sources : :class:`pyuvsim.simsetup.SkyModelData`
+    sources : :class:`pyradiosky.SkyModel`
         The sources to include in the visibility.
     time : :class:`astropy.time.Time`
         Time at which to calculate the visibility.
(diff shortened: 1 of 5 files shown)


wps2n:
@bhazelton I got a memory use improvement on Spectral vs Full and was able to actually run my spectral GSM for 500 freqs that I was not able to before! For quick reference:
```
                Full          Spectral
 50 freqs       ~426GB        ~316GB
 500 freqs      ~2.17TB       ~1.03TB
```
One note: when I produced the full GSM, I used `np.linspace` to fill out the freq_array. For some reason, despite setting the same Nfreqs and start and end freqs in the yaml file for pyuvsim, I get a ValueError saying the freq_array of the uv object and the skymodel do not match. I saw that this check was added to this branch, but I'm not sure why I get I end up raising the ValueError since it looks like pyuvsim also uses `np.linspace`. I was able to fix it by inputting a full freq_array to high precision into the yaml file.

bhazelton:
@wps2n That's great to hear!

I'm not sure what's going wrong with the linspace call, any chance there were different endpoint handling between the calls? If not, it's probably a precision issue. 

Interestingly, the check was always happening before, but it was essentially using a pyuvdata check that we can't rely on now because of how we moved the code around. I would have thought it'd respond the same way as the pyuvdata check, so if that's not the case please give me a minimal example of something that causes an error in this PR that passed before so I can figure out what is going wrong.

wps2n (approved):
Looks good

src/pyuvsim/uvsim.py
```diff
@@ -784,49 +780,81 @@ def run_uvdata_uvsim(
...
-            feed_array=beam_list[0].feed_array,
-            calc_cross_pols=bool(exp_npols > 1),
-            x_orientation=beam_list.x_orientation,
-        )
+        if not (
+            (input_uv.Npols == exp_npols)
+            and (input_uv.polarization_array.tolist() == exp_pols.tolist())
+        ):
+            exp_pols_str = uvutils.polnum2str(exp_pols)
+            raise ValueError(f"input_uv must have polarizations: {exp_pols_str}")
```
mkolopanis, line 819: How do we feel about this error message? I feel like our errors are normally a little more verbose/informative. I think I imagined something like "Input UVData object/simulation parameters and beams polarizations do not agree. Input beams have output polarizations: {}, Simulation has expected polarizations {}"

mkolopanis (commented):
I really only have one question. Looks good but one error message feels a little thin to me.

src/pyuvsim/uvsim.py
bhazelton: done.

mkolopanis (approved):

PR #547, opened 2025-05-06
Fix bug in np.linspace call for frequency and time parameters

bhazelton:
## Description

This fixes a bug found by @wps2n in the np.linspace call in the code that generates frequency and time arrays from input parameters. It is also a significant refactor to reduce code duplication and improve readability.

**This does require updating reference sims 1.2 and 1.3**

## Types of changes

- [ ] Version change

## Checklist:

For all pull requests:

Bug fix checklist:

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -17,7 +17,7 @@ repos:
           args: ['--fix=no']
 
   - repo: https://github.com/astral-sh/ruff-pre-commit
-    rev: v0.11.9
+    rev: v0.11.10
     hooks:
       - id: ruff
       - id: ruff-format


bhazelton:
@wps2n I think this should fix the issue you reported on #540

bhazelton:
@burdorfmitchell This PR is going to require updating the 1.2 & 1.3 reference sims. What's the process for that?

burdorfmitchell:
@bhazelton Sorry for the late reply. 

The process would be to run the new simulations with this updated branch locally, then to upload the reference simulation output to the Brown Digital Repository in the [pyuvsim historical reference simulations collection](https://repository.library.brown.edu/studio/collections/bdr:wte2qah8/) with the same naming conventions as the old simulations. Currently only I have upload capability to the BDR so would need access to the simulation output. I think adding more people to the BDR with upload privilege would be relatively easy but there might be some complications with Brown authentication.

I checked the Run Compare Post Benchmark printed output and reviewed the test_run_ref code and I believe it is just that the comparison tols (rtol and atol) set in pyuvsim [here](https://github.com/RadioAstronomySoftwareGroup/pyuvsim/blob/main/tests/test_run_ref.py#L211) do not match the tols in the equality check in pyuvdata (I believe that has rtol=1e-5 and atol=1e-8, whereas we have rtol=1e-7, atol=0). I guess a temporary bandaid while all the reference simulation changes which are occurring happen could just be to lower the tolerances again as the current simulations will in theory be replaced in about a month.

I will be back in action by the end of the week and can then update the reference simulations on the BDR (finishing up academic year still). If that needs to happen sooner, let me know and I can get that done.

bhazelton:
@burdorfmitchell Thanks for the info, I think it's totally fine to wait until you're done with your academic year. I think we should talk about what tolerance we want to have on these runs. If we want it lower it's easy to update the tolerances on the data array before running the comparison.

bhazelton:
The reference sims have been updated, so those jobs now pass.

jpober (approved):

PR #558, opened 2025-06-06
Test on python 3.13, drop testing on python 3.10

bhazelton:
## Description

We should be testing against 3.13. I dropped 3.10 to not increase the number of versions we're testing against and because we've already dropped it for pyuvdata.

Also updated a few minimum dependency versions:
- scipy >= 1.9
- mpi4py >= 3.1.3

None of these are bleeding edge, they're all fairly old.

NOTE: this branches off of #557, so that must be merged first. This is because the old versions of openmpi are not available for python 3.13.

## Types of changes

- [ ] Version change

## Checklist:

For all pull requests:

- [x] If required or optional dependencies have changed (including version numbers), I have updated the readme to reflect this.

### DIFF
--- a/.github/workflows/testsuite.yaml
+++ b/.github/workflows/testsuite.yaml
@@ -85,7 +85,7 @@ jobs:
   # do this on min_deps because it's hard to get mpi4py to install from pip
   min_deps:
     env:
-      PYTHON: "3.13"
+      PYTHON: "3.12"
     name: Min Deps Testing
     runs-on: ubuntu-latest
     steps:
--- a/.github/workflows/testsuite.yaml
+++ b/.github/workflows/testsuite.yaml
@@ -21,17 +21,17 @@ jobs:
       fail-fast: false
       matrix:
         os: [ubuntu-latest, macos-latest]
-        python-version: ["3.10", "3.11", "3.12"]
+        python-version: ["3.11", "3.12", "3.13"]
         env_name: [pyuvsim_tests_openmpi]
         include:
           - env_name: pyuvsim_tests_mpich
-            python-version: "3.11"
+            python-version: "3.13"
             os: ubuntu-latest
           - env_name: pyuvsim_tests_mpich
-            python-version: "3.11"
+            python-version: "3.13"
             os: macos-latest
           - env_name: pyuvsim_tests_windows
-            python-version: "3.12"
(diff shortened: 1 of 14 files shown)


src/pyuvsim/mpi.py
```diff
@@ -51,11 +51,13 @@ def start_mpi(block_nonroot_stdout=True):
         Redirect stdout on nonzero ranks to /dev/null, for cleaner output.
 
     """
+    do_once = False
     global world_comm, node_comm, rank_comm, rank, Npus
-    if not MPI.Is_initialized():
+    if not MPI.Is_initialized():  # pragma: no cover
```
bhazelton, line 56: This isn't covered because we use `mpi-pytest` to run our tests which initializes MPI before we get here.

.github/workflows/testsuite.yaml
```diff
@@ -210,7 +210,7 @@ jobs:
 
       - name: Run Tests
         run: |
-          python -m pytest -n auto -W error --cov=pyuvsim --cov-config=.coveragerc --cov-report xml:./coverage.xml --junitxml=test-reports/xunit.xml
+          pytest -W error --cov=pyuvsim --cov-config=.coveragerc --cov-report xml:./coverage.xml --junitxml=test-reports/xunit.xml
```
steven-murray, line 213: Why did `-n auto` need to be removed?

tests/test_run.py
```diff
@@ -43,6 +43,10 @@ def goto_tempdir(tmpdir):
 @pytest.mark.parametrize(
     "paramfile", ["param_1time_1src_testcat.yaml", "param_1time_1src_testvot.yaml"]
 )
+@pytest.mark.skipif(
+    not pytest.pyuvsim_can_parallel,
+    reason="mpi-pytest is not installed. Cannot run parallel tests.",
+)
```
steven-murray, line 49: Does `run_uvsim` require MPI? It can't run in an mpi-less mode?

steven-murray (commented):
I think this looks good, though this is the kind of PR where mistakes are more likely to happen because you forgot to change something, rather than you changing something incorrectly, and that's much easier to miss in the review. But I can't think of anything!

Anyway, just the two short questions.

steven-murray (approved):

Issue #469, opened 2024-06-04
Convert scripts to entry-point functions, the preferred option with pyproject.toml


bhazelton:
This means removing the scripts as stand alone files and refactoring them into functions that can be plumbed as entry-points in the pyproject.toml.

mkolopanis:
a good resource for this https://setuptools.pypa.io/en/latest/userguide/entry_point.html

It does basically end up with having a function that does something testable, and a thin wrapper around that function which sets up argument parsing

Issue #437, opened 2023-10-12
HERA beam in reference sims


jpober:
Our reference sims currently use the Fagnoni UVBeam as our example of a UVBeam.  We should probably switch to something non-proprietary... do we have another UVBeam that makes sense to use?

PR #569, opened 2025-07-09
Implementation of new first generation reference simulations.

### LINKED ISSUE #419: Google Drive link for reference sims
There appear to be two links out there: 

- The one in our README: https://drive.google.com/drive/folders/0B0DC3mpkVdq6fjhZTWdsV2pBQXNKNHVWeExRd3gzdFN4WXp3cDVLaTlZZHlrVmZ0eFNtOHM?resourcekey=0-AOKBEf0D_nO0o8uOO-g-QQ&usp=share_link

- One from Daniya, which appears more recent: https://drive.google.com/drive/folders/14hH-zBhHGddVacc0ncqRWq7ofhGLWfND?usp=share_link

There is no definitive answer which is right.  (And some of the other issues with the reference sims are not solved merely by switching links.)

### LINKED ISSUE #421: Reference sim 1.2_zenith changes
New versions of reference sim 1.2_zenith (regardless of beam) do not match the old ones (regardless of which Google Drive link).  There are two source catalogs that could be used for this reference sim in the repo, too, but neither give visibilities that match the Google Drive either.

### LINKED ISSUE #460: Change file type in reference sim docs
The READMEs for the reference simulations list the expected files as .uvfits.  We changed the default output to .uvh5, so we should update those docs.

### LINKED ISSUE #499: Allow relative path for uvbeam specification
It would useful to allow a relative beam_path specification for the uvbeam file in the simulation config yamls. We currently use the absolute path, which is understandable because the user is required to download the uvbeam file separately. A relative path would require less file editing for each user, and having an absolute path such as `/users/dseitova/data/dseitova/HERA_NicCST_fullfreq.uvbeam` as the default feels wrong (we have a better placeholder for the 2nd generation simulations but that one might also be deceptive).

### LINKED ISSUE #550: Multi beam reference simulation
I don't think our reference simulations (even in the new versions under development) ever use more than one beam.  We probably want that?

### LINKED ISSUE #556: Lunar reference sim
I think we should have at least one reference simulation featuring our lunar simulation capabilities.  We may want a further suite of lunar simulations, but one that's part of the official reference simulation suite seems good.

burdorfmitchell:
## Description

A full overhaul of the first generation reference simulations. All first generation reference simulation yamls removed and replaced with a new set of 8 reference simulations. The implementation of new reference simulations is documented in two new Memos: "Documentation of New Reference Simulations" and "Reference Simulation Renovation". The scripts to download necessary data files and reference simulations have been converted to entry points in cli.py -- `download_data_files` to download the 3 large data files to run the reference simulations, and `download_ref_sims` to download reference simulations stored on the Brown Digital Repository. Testing and docs have been updated correspondingly. astropy is used to handle downloading and caching large data files -- PR #563 and [this PR in pyuvdata](https://github.com/RadioAstronomySoftwareGroup/pyuvdata/pull/1583) are used to allow the reference simulation yamls to target cached files by their urls, allowing for individuals to download the files with the command line tool and immediately run the reference simulations without having to move data files to specific path locations or edit the yaml files with absolute paths.

## Motivation and Context

closes #556, closes #550, doesn't close but kind of resolves #499, works with #469, closes (for first generation only but could easily be edited to cover second generation for this issue) #437, closes #421, closes #419, closes #460, 

## Types of changes

- [x] Reference simulation update or replacement
- [ ] Version change

## Checklist:

For all pull requests:

New feature checklist:

Breaking change checklist:
- [ ] I have updated the docstrings associated with my change using the [numpy docstring format](https://numpydoc.readthedocs.io/en/latest/format.html).
- [ ] I have updated the documentation to reflect my changes (if appropriate).
- [ ] My change includes backwards compatibility and deprecation warnings (if possible).
- [ ] I have added tests to cover my changes.

Reference simulation update or replacement:
- [x] If this is a new reference simulation or if the settings files have changed, I have added all the relevant files to the reference_simulations folder.
- [ ] If this is a new reference simulation I have fully documented it in a memo in docs and updated any other docs as appropriate.

Version change checklist:
- [ ] I have updated the [CHANGELOG](https://github.com/RadioAstronomySoftwareGroup/pyuvsim/blob/main/CHANGELOG.md) to put all the unreleased changes under the new version (leaving the unreleased section empty).
- [ ] I have run the appropriate tests for this level of version change (described in the readme under versioning).


.github/workflows/testsuite.yaml
```diff
@@ -220,3 +220,53 @@ jobs:
           file: ./coverage.xml #optional
           env_vars: OS,PYTHON
           fail_ci_if_error: true
+
+  download_test:
```
bhazelton, line 224: I don't quite understand what this job is adding that is not covered by other jobs?

CHANGELOG.md
```diff
@@ -14,6 +14,9 @@ feed_angle, mount_type) which will be released in pyuvdata 3.2
 - Updated minimum dependency versions: python>=3.11, pyuvdata>=3.1.2, pyradiosky>=1.1.0, scipy>=1.9
 - Updated minimum optional dependency versions: mpi4py>=3.1.3
 - Only import lunarsky if needed.
+- Implemented a new set of first generation reference simulations and updated documentation. Moved
+data and reference simulation download scripts to entry points which download files to the pyuvsim
+cache using astropy.
```
bhazelton, line 19: This should go at the top of this section, it should be in reverse chronological order.

reference_simulations/first_generation/telescope_config/mwa_config_uvbeam.yaml
```diff
@@ -0,0 +1,8 @@
+beam_paths:
+  0: !UVBeam
+      filename: http://ws.mwatelescope.org/static/mwa_full_embedded_element_pattern.h5
+      pixels_per_deg: 1
+      file_type: mwa_beam
```
bhazelton, line 5: I'd suggest adding `mount_type: phased` to avoid some deprecation warnings that will turn into errors in the future.

src/pyuvsim/data/mwa_128T_layout.csv
```diff
@@ -0,0 +1,130 @@
+Name	Number	BeamID	E	        N	        U
```
bhazelton, line 1: Seems like we have this file in the repo twice? here and in the reference sims folder. Is that what we want?

bhazelton: I think that's also true of a bunch of the following files here. I worry that if we have them in twice they can get out of sync...

tests/test_run_ref.py
```diff
@@ -213,85 +158,71 @@ def compare_uvh5(uv_ref, uv_new):
...
+        run_uvsim,
+        args=[yaml_filepath],
+        kwargs={"return_uv": True},
+        setup=None,
+        rounds=2,
+        warmup_rounds=0,
+        iterations=1,
+    )
+    # else:
+    #    uv_new = benchmark(run_uvsim, yaml_filepath, return_uv=True)
```
bhazelton, line 225: Is this going to change before merging?

.github/workflows/testsuite.yaml
bhazelton: Ah, ok. I had missed the comment above that test that explained how it was prevented from being run on all the other jobs.

src/pyuvsim/data/mwa_128T_layout.csv
bhazelton: We talked on the telecon and decided that we should put them in the data folder and just point to where they are (with urls) in the reference sim documentation. And we should update the download scripts to get them from the right place. The idea long term is that we'll move the test and ref sim data to another repo and use pooch to get them.

reference_simulations/catalog_files/MOON.txt
```diff
@@ -0,0 +1,150 @@
+SOURCE_ID	RA_ICRS [deg]	Dec_ICRS [deg]	Flux [Jy]	Frequency [Hz]
```
bhazelton, line 1: Looks like there's a few files still here in the reference_simulations folder that are duplicated in the src/pyuvsim/data folder.

src/pyuvsim/cli.py
```diff
@@ -364,7 +375,7 @@ def create_text_catalog(
     source_coord = SkyCoord(
         alt=Angle(alts, unit=u.deg),
         az=Angle(azs, unit=u.deg),
-        obstime=time,
+        obstime=time.iso,
```
bhazelton, line 378: This is interesting, why are you converting it to iso string here rather than leaving it as an astropy Time object? What does it fix?

burdorfmitchell: The skyh5 file is written successfully, but the written skyh5 file cannot be read. This appears to be because in pyradiosky skymodel.py line 287 it tries to use the parsed time value to create a time object, and jd does not work without additional input (e.g. Time("2460000.0", format="jd") works but the code is only calling Time("2460000.0") which fails).

bhazelton: ok, I pushed those changes.

PR #572, opened 2025-07-14
Use mpi pytest

mkolopanis:
changes the parallel decorator to use `mpi-pytest` package instead of our own homebrew one.

I'm pretty sure new warnings are being issued now from mpi-pytest and that breaks the warning test. maybe we just need to ignore them?


mkolopanis:
but interestingly, the codecov checks passed before the python xdist was re-added, could remove that from the warnings environment too?

bhazelton:
> @bhazelton I think we're missing the mpi inits because of mpi-pytest at this point. It makes a call to `import mpi4py` in its stack which does call the mpi initialization command.
> 
> I personally would be okay taking the 3 line hit but we can try to make a test suite that doesn't use mpi-pytest but still has mpi? 🤔

I'm fine with just slapping a `"# pragmatic: no cover"` on that block.

mkolopanis:
oh right we stopped invoking it :thinking: well why did it work before but not now.

PR #617, opened 2026-02-19
Unpin freetype in Windows builds

### LINKED ISSUE #595: unpin freetype in windows ci conda yaml
In #589 I pinned `free_type<2.14` in `ci/pyuvsim_tests_windows.yaml` because version 2.14, which had just been released hours before on conda, was causing "access violation" errors on Windows tests. I found an old issue where that had been caused by freetype in the past (https://github.com/matplotlib/matplotlib/issues/22006) and found that a new version had just been released, so I decided to try pinning it and that solved the problem. The error I was seeing was nearly identical to the error described in that old issue.

We should test unpinning freetype once a new version is out.

bhazelton:
## Description

Remove the pin on freetype in the windows CI conda yaml.

## Motivation and Context

closes #595 

## Types of changes

- [ ] Version change

## Checklist:

For all pull requests:

- [x] If required or optional dependencies have changed (including version numbers), I have updated the readme to reflect this.

### DIFF
--- a/ci/pyuvsim_tests_windows.yaml
+++ b/ci/pyuvsim_tests_windows.yaml
@@ -6,7 +6,6 @@ dependencies:
   - astropy-healpix>=1.0.2
   - coverage
   - line_profiler
-  - freetype<2.14
   - matplotlib>=3.6.*
   - mpi4py>=3.1.3
   - numpy>=1.23


burdorfmitchell (approved):
Sorry for taking so long to get to this. It looks good to me. I have confirmed that the issues with the reference simulation regression testing are again due to astropy-iers and updated the Brown Digital Repository so the tests should now pass.

bhazelton:
@burdorfmitchell Looks like both this PR and #621 are seeing a slight speed regression in the reference sims, but they are on different sims! Also, neither changes the actual code in any way, they're both just CI setup changes that should have no effect on the jobs that run the reference sims. Thoughts?

bhazelton:
I re-ran all the reference sims. Before the rerun, 1.1 and 1.4 were identified as running slower. 1.1 had a ratio of 1.22 and 1.4 had a ratio of 1.3.

After the re-run, only 1.8 is identified as running slower, with a ratio of 1.25.

So I think this is something to do with the GitHub runners they are ending up on rather than the actual code.

I'm re-running just 1.8 now to see what happens.

burdorfmitchell:
I also believe that is the case because many simulations have run significantly faster as well without clear explanation, the “regressions” are inconsistent as you noted, and I found a post about GitHub migrating its Ubuntu images for the image versions used in these actions. It is a little awkward because a slowdown of ~30% doesn’t feel like something to ignore but we could bump the regression tolerance to around that. We will still have the plot of the benchmarks over time to compare to in the case a slowdown is real and persistent.

burdorfmitchell (approved):

PR #630, opened 2026-07-15
Retry pytest steps in actions to handle intermittent errors

bhazelton:
## Description

Intermittent errors in GitHub actions have been driving us crazy, so this adds automatic retries of the steps that usually error.

There is an existing action that can be used for this, but it doesn't work with conda because there's no way to use the right shell. So I used that only on the job that uses pypi. In the conda jobs I just manually check if the step passed and re-run it if it didn't.

I also consolidated the warnings test and min_versions test into the top level testing job using matrix includes and simplified the windows os handling the same way.

update:  It looks like if the first attempt fails on the weird mpi error so do the retries. Which implies that the mpi error is related to the machine or the installed versions. There are other intermittent errors, so I think this is still helpful, but it doesn't seem to fix all the issues.

## Types of changes

- [ ] Version change

## Checklist:

For all pull requests:

### DIFF
--- a/.github/workflows/compare-post-benchmark.yaml
+++ b/.github/workflows/compare-post-benchmark.yaml
@@ -30,6 +30,7 @@ jobs:
     env:
       ENV_NAME: pyuvsim_tests_refsim
       PYTHON: "3.14"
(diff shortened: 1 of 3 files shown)


steven-murray (dismissed):
It's hacky, but I don't know a better way

burdorfmitchell (approved):
This approach seems fine to me.

PR #648, opened 2026-09-20
Pass existent UVData baselines through obsparam select.bls

### LINKED ISSUE #508: uvdata_to_telescope_config doesn't pass baseline subsets
If you load a UVData object, run select on it to eliminate some baselines, and then create a yaml file, all the baselines still get simulated.  One solution would be for this method check the baseline_array and then pass the list of actually existent baselines to the select keyword in the yaml file.  There may be other things that don't get passed to the yaml file too (frequencies seems to work, at least).

shaneraphel:
## Description

`uvdata_to_config_file` already copies frequency and time from the input `UVData` arrays, but it never wrote `select.bls`. After `UVData.select` drops some baselines, the obsparam yaml still has no baseline list, so `initialize_uvdata_from_params` builds the full combinatorial array and the simulation is not the subset that was selected.

The writer now records the existent `(ant1, ant2)` pairs from `ant_1_array` / `ant_2_array` on `select.bls`, the same key initialize already pops before `UVData.new`. Telescope yaml (name, location, layout, spoofed beams), freq/time dicts, and `subselect` are unchanged.

A roundtrip test keeps one triangle baseline through write then initialize.

## Motivation and Context

closes #508

## Types of changes

## Checklist:

For all pull requests:

New feature checklist:


bhazelton:
I don't think we want to do this. It would write out all the antenna pairs in the data every time. That would make the yaml massive and hard to read/edit for people, which is the point of the yaml. We need to discuss an appropriate approach to this problem, moving that discussion back to the issue.

shaneraphel:
Agreed. Writing every existent `(ant1, ant2)` pair into `select.bls` on every call makes the yaml too large to read or edit, which is the point of the file. I will not pursue that approach on this pull request.

The underlying problem in #508 is still real: after `UVData.select` drops baselines, `uvdata_to_config_file` keeps frequency and time but not the baseline subset, so `initialize_uvdata_from_params` rebuilds the full combinatorial array. Discussion of a compact way to record that subset belongs on the issue, not here.

shaneraphel:
Closing so the baseline-subset discussion stays on #508. The full select.bls dump is not the approach we should land.

